PR #2, opened 2016-12-02
sam has opinions

tychoish:


.evergreen/config.yml
```diff
@@ -314,10 +285,10 @@ functions:
   "upload test results":
     - command: attach.results
       params:
-        file_location: "${PROJECT_DIRECTORY}/test-results.json"
+        file_location: "test-results.json"
     - command: attach.results
       params:
-        file_location: "${DRIVERS_TOOLS}/results.json"
```
bjori: I don't think this will work, these are two different directories these files are in

.evergreen/config.yml
```diff
@@ -36,60 +36,51 @@ functions:
     # Make an evergreen exapanstion file with dynamic values
     - command: shell.exec
       params:
-        working_dir: "src"
+        shell: bash
         script: |
-           echo "" > expansion.yml
-
+           pushd src
```
bjori: its not available on all the distros. this was the first thing I did too, but got errors :(

.evergreen/config.yml
```diff
@@ -36,60 +36,51 @@ functions:
...
-                 # Python has problems with unix style paths in cygwin. Must use c:\\ paths
-                 export DRIVERS_TOOLS="c:/drivers-tools"
-                 ;;
-              *)
-                 export DRIVERS_TOOLS="$PWD/../drivers-tools"
-                 ;;
-           esac
+
+           export TASK_DIR=`pwd`
+           if [ "Windows_NT" == "$OS" ]; then
```
bjori: is $OS evergreen variable ?

.evergreen/config.yml
```diff
@@ -150,52 +126,48 @@ functions:
...
 
   "exec compile script" :
     - command: shell.exec
       type: test
       params:
-        working_dir: "src"
         script: |
           ${PREPARE_SHELL}
-          [ -f ${PROJECT_DIRECTORY}/${file} ] && BUILDTOOL="${buildtool}" sh ${PROJECT_DIRECTORY}/${file} || echo "${PROJECT_DIRECTORY}/${file} not available, skipping"
+          [ -f ${file} ] && BUILDTOOL="${buildtool}" sh ${file} || echo "${file} not available, skipping"
```
bjori: Are you sure this works?

The git checkout is into `src` and we want to execute the file from that working directory so that relative paths in the project directory work successfully.

.evergreen/config.yml
```diff
@@ -150,52 +126,48 @@ functions:
...
     - command: shell.exec
       params:
         silent: true
         script: |
+          ${PREPARE_SHELL}
+
            export AWS_ACCESS_KEY_ID=${aws_key}
            export AWS_SECRET_ACCESS_KEY=${aws_secret}
-           aws s3 cp ${PROJECT_DIRECTORY}/doc/html s3://mciuploads/${UPLOAD_BUCKET}/docs/${CURRENT_VERSION} --recursive --acl public-read --region us-east-1
+           aws s3 cp doc/html s3://mciuploads/${project}/docs/${CURRENT_VERSION} --recursive --acl public-read --region us-east-1
```
bjori: wouldn't this need to be `src/doc/html` ?

.evergreen/config.yml
tychoish: Being able to change the shell to bash (which is a new feature) makes this work, I'm pretty sure.

.evergreen/config.yml
tychoish: I think it's populated in cygwin bash; this is pulled from the evergreen self-tests, and seems to be the prevailing idiom.

.evergreen/config.yml
tychoish: Can we just glob? 

Why are we writing test output to two different places?

tychoish:
updated with some fixes, patching here: https://evergreen.mongodb.com/version/58497e493ff122540a00315d

bjori:
Merged parts of this pr in https://github.com/mongodb-labs/drivers-evergreen-tools/commit/022b03e8a19ac6a8a41166c043d16e29f8245085

I realized we can't use the `module` approach as that wouldn't work for patchbuilds.

PR #6, opened 2017-01-20
Force LF line endings with .gitattributes

jrassi:


.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Force shell scripts to always have LF endings, even on Windows.
+*.sh text eol=lf
```
bjori: I don't think .gitattributes are applied for evergreen patches, so I don't think this won't work when adding a file in a patch build, fyi.

bjori:
Sorry, I missed this PR. FYI; I'd be careful to rely on this, and instead script it like is done on https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/config.yml#L353

That being said, I'll merge this as it does make sense, thanks!

PR #8, opened 2017-01-26
Fallback to generic Linux x86_64 builds (without SSL) when no platform specific link is available

ShaneHarvey:
This will let drivers choose to test Linux x86_64 variants with all MongoDB versions (without SSL). For example, the Python driver wants to test all versions of Python against all version of the server on Ubuntu 16.04 (x86_64). When using the generic Linux builds for MongoDB 2.4, 2.6, and 3.0 we'll simply skip the SSL tests.


bjori:
nice. Thanks!

PR #10, opened 2017-01-27
Factor out stopping mongo-orchestration

xdg:
This splits stopping orchestration out of the `cleanup` function and puts
the actual work into a separate script.  This makes it more reusable by
Evergreen configs that don't exactly duplicate driver-tools.

Patch build is looking good: https://evergreen.mongodb.com/version/588b57893ff1224075014cebo


bjori:
Makes sense, thanks!

PR #12, opened 2017-01-30
Fix "install dependencies" function definition

ShaneHarvey:
Before this change the "install dependencies" pre hook was silently skipped, see https://jira.mongodb.org/browse/EVG-763.


bjori:
Nice debugging. Thanks!

PR #20, opened 2017-04-29
Don't start MongoDB with --nohttpinterface

ajdavis:
Removed in MongoDB 3.5.7+, SERVER-29000.


ajdavis:
@bjori and @ShaneHarvey, look reasonable? I don't know why it was important in the past to start MongoDB with --nohttpinterface.

bjori:
Interesting. I wonder if there were some security concerns at the time. I can't see any issues it could create today. 
This lgtm

PR #23, opened 2017-10-26
Start shards as single member replica sets.

ShaneHarvey:
ChangeStreams and retryable writes require replica sets.


ShaneHarvey:
Waiting on patch to succeed: https://evergreen.mongodb.com/version/59f27233e3c3310b8f003a83

PR #44, opened 2018-05-24
Use Python 3 from the toolchain when available

ShaneHarvey:


ShaneHarvey:
Patch: https://evergreen.mongodb.com/version/5b07074a2a60ed13d3cae377

ShaneHarvey:
Note, Amazon linux is still failing because of https://jira.mongodb.org/browse/BUILD-5581

PR #60, opened 2019-02-05
Update download script to use new names for latest on osx and windows

samantharitter:


behackett:
@samantharitter - you can merge this now.

PR #62, opened 2019-03-05
PYTHON-1644 Ensure primary starts on port 27017

ShaneHarvey:
Fixes [PYTHON-1644](https://jira.mongodb.org/browse/PYTHON-1644): https://evergreen.mongodb.com/version/5c7ee3e532f4176c5c1f23c1

The remaining test failures are from a server change which added the `keyPattern` and `keyValue` fields to writeErrors.

### DIFF
--- a/.evergreen/orchestration/configs/replica_sets/auth-ssl.json
+++ b/.evergreen/orchestration/configs/replica_sets/auth-ssl.json
@@ -13,6 +13,7 @@
         "port": 27017
       },
       "rsParams": {
+        "priority": 1,
         "tags": {
           "ordinal": "one",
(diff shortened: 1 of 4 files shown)


ShaneHarvey:
[According the docs the default priority is 1 ](https://docs.mongodb.com/manual/tutorial/adjust-replica-set-member-priority/#considerations)so we need to use something higher, like 10.

ShaneHarvey:
Reverted this change because various pymongo replica set tests started to fail with not master or socket closed errors. Will investigate later but my initial theory is that MO doesn't verify that the highest priority node is actually the primary so later an election is triggered during the test suite.

PR #67, opened 2019-06-26
Fix ubuntu1604 arm64 download paths

prashantmital:
This PR updates the download paths for the latest and 4.2 server versions for Ubuntu 16.04 arm64. 
I have sought confirmation from the server build team that the hosting paths have been modified as indicated in this PR and will update this PR once I hear back.


prashantmital:
Waiting for write-access before I can merge this.

PR #73, opened 2019-09-09
Update with download links for Debian 10

rcsanchez97:


.evergreen/download-mongodb.sh
```diff
@@ -240,6 +240,17 @@ get_mongodb_download_url_for ()
              MONGODB_26=""
              MONGODB_24=""
       ;;
+      linux-debian-10*)
+         MONGODB_LATEST="http://downloads.10gen.com/linux/mongodb-linux-x86_64-enterprise-debian92-latest.tgz"
```
ShaneHarvey: Do the download links for Debian 10 really have `-debian92-`?

rcsanchez97: No.  There are no debian10 download links yet (probable because the AMI just came available in Evergreen).  The debian92 builds work on debian10 and once debian10 builds are available I will update again to use those.

ShaneHarvey: Thanks for explaining that!

rcsanchez97:
@ShaneHarvey It looks like the server team has started publishing a latest tarball for debian10.  I have updated to use that.  There are no debian10 downloads for prior release versions.

PR #75, opened 2019-09-24
rename link to ca.pem with new subject hash value

bazile-clyde:


bazile-clyde:
@behackett just realized this was still opened. Would you like to review it or would rather I defer it someone else?

behackett:
I didn't realize I was on this review. What changed about the CA file?

bazile-clyde:
OpenSSL needs a hashed file name of our server certificate which can be obtained by using the cmd: `openssl x509 -hash -noout -in ca.pem`. Since the server cert was updated [DRIVERS-575](https://jira.mongodb.org/browse/DRIVERS-575) the hash of this file also needs to be updated.

PR #79, opened 2020-01-06
Add OCSP testing suite

vincentkam:
Needed as part of https://jira.mongodb.org/browse/WRITING-4242

The certs were generated using https://github.com/mongodb/mongo/blob/master/jstests/ssl/x509/mkcert.py
~~https://github.com/mongodb/mongo/blob/master/jstests/ssl/x509/certs.yml~~


.evergreen/ocsp/mock_ocsp_revoked.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env sh
+# Use the CA as the OCSP responder (see https://community.letsencrypt.org/t/unable-to-verify-ocsp-response/7264)
+python3 ocsp_mock.py \
+  --ca_file ca_ocsp.pem \
+  --ocsp_responder_cert ca_ocsp.crt \
```
vincentkam: Per the comment above, I'm using the CA as the OCSP responder, which according to the above link, is also what Let's Encrypt does. Doing so simplifies the certificate chain and seems to reduce the amount of setup needed for Java at the very least (obviating the need to explicitly trust a separate OCSP responder certificate).

.evergreen/ocsp/ocsp_mock.py
```diff
@@ -0,0 +1,46 @@
+#! /usr/bin/env python3
```
vincentkam: Taken from https://github.com/mongodb/mongo/blob/master/jstests/ocsp/lib/ocsp_mock.py.

vincentkam:
Ready for review 😸

.evergreen/ocsp/mock_ocsp_revoked.sh
```diff
@@ -0,0 +1,9 @@
...
+  --ocsp_responder_key ca_ocsp.key \
+   -p 8100 \
```
jmikola: Where does port 8100 come from? Curious if this would conflict with anything else drivers might run and whether it may need to be configurable. I suppose we could have optional shell script arguments and default to whatever value we like.

http://wiki.cacert.org/OcspResponder suggests 2560 (seemingly reference to the RFC).

vincentkam: The port is pseudo-configurable via the -p parameter. If we updated the port, we'd need to update the certs. (see https://github.com/mongodb/mongo/blob/master/jstests/ssl/x509/certs.yml#L359 for where we configure this).

 Ah, interesting find on the suggested port! As a counterpoint though, Let's Encrypt appears to be using port 80 (OCSP - URI:http://isrg.trustid.ocsp.identrust.com, as show by the output of `openssl s_client -host valid-isrgrootx1.letsencrypt.org -port 443 -prexit -showcerts`)

I'd recommend leaving the port the same for now as it's simpler, although if a conflict does exist, we can change it.

.evergreen/ocsp/ocsp_mock.py
vincentkam: Added the link as a comment to both python files!

.evergreen/ocsp/mock_ocsp_revoked.sh
jmikola: Fine keeping 8100 and reconsidering if/when a conflict comes up. I didn't realize that it was hard-coded into the certs, but I suppose that makes sense now that I think about it.

.evergreen/ocsp/ocsp_mock.py
jmikola: Did you push the changes? I don't see a link in the current diff. Feel free to resolve this once changes are reflected in the PR.

vincentkam: Ah, sorry, was on a hotspot when I was pushing, something went wrong. Now it's truly pushed!

vincentkam:
@jyemin That's true, I'll add an extra configuration

.evergreen/orchestration/configs/servers/basic-tls-ocsp-mustStaple.json
```diff
@@ -0,0 +1,17 @@
...
+   "id" : "standalonenoauthssl",
+   "name": "mongod",
+   "procParams": {
+      "ipv6": true,
+      "bind_ip": "127.0.0.1,::1",
+      "logappend": true,
+      "journal": true,
+      "port": 27017
+   },
+   "sslParams": {
```
vincentkam: I believe using the older parameter names is needed for maximum backwards compatibility.

vincentkam:
I've regenerated the certs using tweaked versions of the aforementioned python script and yml file: I'll create a pull request for those later. Ready for re-review!

behackett:
IIRC, you were going to add instructions?

vincentkam:
@behackett  Apologies! I can't find a comment in the spec regarding adding instructions, and I can't remember what instructions were to be added. Although fwiw, instructions for how to use the mock responder included in the test plan can be found here (https://docs.google.com/document/d/1q-YEvyjbaKgtAKhZYvyGc20SyZjPfLQ0GJhBfvM4zUk/edit#heading=h.754fl0etewfs). Unless we wanted to add instructions how to generate the certificates?

.evergreen/orchestration/configs/servers/basic-tls-ocsp-mustStaple.json
jmikola: I believe that's correct, as the "tls" configuration API was introduced in [4.2](https://docs.mongodb.com/manual/tutorial/configure-ssl/#procedures-using-net-ssl-settings).

jmikola:
> Unless we wanted to add instructions how to generate the certificates?

If the certs in this PR were generated from scratch and not simply pulled from the server, then I think it'd be worth documenting the steps used to produce them in a README to include alongside them in this PR.

behackett:
Sorry @vincentkam, my comment was based on a discussion we previously had. I'm realizing the scripts document what we previously discussed. Carry on. :-)

vincentkam:
I've created a PR with the changes to the generation script needed to generate the certs here: https://github.com/mongodb/mongo/pull/1345
I'm going to add the yml file I used to generate the certificates to this PR as well as a README

vincentkam:
`certs.yml` and a `README.md` added. 

Ready for review!

jmikola:
@vincentkam FYI, `README.md` is included in the PR but it's completely empty.

vincentkam:
@jmikola Fixed, sorry!

.evergreen/ocsp/certs.yml
```diff
@@ -0,0 +1,68 @@
+global:
+  # All subject names will have these elements automatically,
+  # unless `explicit_subject: true` is specified.
+  output_path: '../drivers-evergreen-tools/.evergreen/ocsp/'
```
jmikola: I suggest adding some comment here that refers back to the `README.md` file, even if it's just "See README.md if customizing this path" or something to that effect.

.evergreen/ocsp/README.md
```diff
@@ -0,0 +1,9 @@
+# Generating test certificates
+The test certificates here were generating using the server team's `mkcert.py` tool, which can be found here: https://github.com/mongodb/mongo/blob/master/jstests/ssl/x509/mkcert.py.
+
+In order to generate a fresh set of certificates, the following command should be used from the root of the `mongo` repository (taking into account the location of the `certs.yml` file in this directory):
+` python3 jstests/ssl/x509/mkcert.py --config ../drivers-evergreen-tools/.evergreen/ocsp/certs.yml`
+
+The certificates will be output into the folder specified in `certs.yml`. (The default configuration assumes that the `mongo` repository and the `driver-evergreen-tools` repository have the same parent directory.
+
+The final step is to split the `ca_ocsp.pem` file, which contains both the private key and the public certificate, into two files. `ca_ocsp.crt` should contain the public certificate, and `ca_ocsp.key` should contain the private certificate.
```
jmikola: Suggested changes:

```
# Generating Test Certificates

The test certificates here were generating using the server team's [`mkcert.py`](https://github.com/mongodb/mongo/blob/master/jstests/ssl/x509/mkcert.py) tool.

In order to generate a fresh set of certificates, clone the [`mongo` repository](https://github.com/mongodb/mongo) and run the following command from the root of the `mongo` repository:

'''
python3 jstests/ssl/x509/mkcert.py --config ../drivers-evergreen-tools/.evergreen/ocsp/certs.yml
'''

The certificates will be output into the folder specified by the `global.output_path` option in `certs.yml`, which defaults to the same directory as this README file and `certs.yml`. The default configuration also assumes that the `mongo` repository and the `driver-evergreen-tools` repository have the same parent directory.

The final step is to split the `ca_ocsp.pem` file, which contains both the private key and the public certificate, into two files. `ca_ocsp.crt` should contain the public certificate, and `ca_ocsp.key` should contain the private certificate.
```

Note: I used `'''` instead of backticks above since I slapped the entire suggestion in a code block. Please convert accordingly if you take this suggestion.

vincentkam: I've taken your suggestions!

vincentkam:
Rebased, squashed, date amended, and pushed!

PR #80, opened 2020-01-21
Enable OCSP support for 4.3.x servers

vincentkam:
OCSP on 4.3.x has been flag gated under a set parameter, so we need to include `--setParameter ocspEnabled=true`

Following the format for setParamter laid out in: https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/orchestration/configs/servers/disableTestCommands.json


vincentkam:
Ready for review :)

PR #87, opened 2020-02-14
Stop using nohup 

ShaneHarvey:
nohup recently started failing on the macos hosts with:
```
nohup: can't detach from console: Inappropriate ioctl for device
``` 

We don't need to use nohup here as evidenced by this passing python patch build: https://evergreen.mongodb.com/version/5e47126a9ccd4e19cba49c37

### DIFF
--- a/.evergreen/start-orchestration.sh
+++ b/.evergreen/start-orchestration.sh
@@ -59,7 +59,7 @@ else
   fi
 fi
 
-set -o pipefail && mongo-orchestration $ORCHESTRATION_ARGUMENTS start < /dev/null | tee $MONGO_ORCHESTRATION_HOME/out.log
+mongo-orchestration $ORCHESTRATION_ARGUMENTS start
 
 ls -la $MONGO_ORCHESTRATION_HOME
 
--- a/.evergreen/start-orchestration.sh
+++ b/.evergreen/start-orchestration.sh
@@ -59,7 +59,7 @@ else
   fi
 fi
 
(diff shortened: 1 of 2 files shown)


.evergreen/start-orchestration.sh
```diff
@@ -59,7 +59,7 @@ else
   fi
 fi
 
-nohup mongo-orchestration $ORCHESTRATION_ARGUMENTS start > $MONGO_ORCHESTRATION_HOME/out.log 2> $MONGO_ORCHESTRATION_HOME/err.log < /dev/null &
+set -o pipefail && mongo-orchestration $ORCHESTRATION_ARGUMENTS start < /dev/null | tee $MONGO_ORCHESTRATION_HOME/out.log
```
ShaneHarvey: This change will make it easier to diagnose `mongo-orchestration start` failures because:
- we now fail fast when `start` fails
- we log the `start` output to stdout in addition to `out.log`
- we don't need to use `&` because `start` launches a daemon process

ShaneHarvey: Turns out we do need to use `&` or else the `start` command hangs. I'm not sure why and I don't want to find out. We also can't use pipefail on RHEL 7.0 (I think we could change these scripts to use `/bin/bash` instead of `/bin/sh` but that might open a whole new can of worms. Instead I changes this back to use `&` but I did add some better logging when MO fails to start (or fails to start a cluster).

ShaneHarvey:
Final passing patch: https://evergreen.mongodb.com/version/5e471ddce3c33116e7803fea

PR #90, opened 2020-02-21
Fix latest download link for windows

divjotarora:
This PR only fixes the link for cygwin*-x86_64 because that's the only set of Windows downloads using downloads.10gen.com (the rest use fastdl.mongodb.org) and it's the case that Go and .NET match in Evergreen. I'm not sure if the other cases have to be updated too.﻿

### DIFF
--- a/.evergreen/download-mongodb.sh
+++ b/.evergreen/download-mongodb.sh
@@ -387,15 +387,16 @@ get_mongodb_download_url_for ()
              MONGODB_24="https://fastdl.mongodb.org/win32/mongodb-win32-i386-${VERSION_24}.zip"
       ;;
       windows64*)
+         # same as cygwin*-86-64
          EXTRACT="/cygdrive/c/Progra~2/7-Zip/7z.exe x"
-         MONGODB_LATEST="https://fastdl.mongodb.org/win32/mongodb-win32-x86_64-2012plus-latest.zip"
-             MONGODB_42=""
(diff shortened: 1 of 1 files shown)


.evergreen/download-mongodb.sh
```diff
@@ -401,7 +401,7 @@ get_mongodb_download_url_for ()
       ;;
       cygwin*-x86_64)
          EXTRACT="/cygdrive/c/Progra~2/7-Zip/7z.exe x"
-         MONGODB_LATEST="http://downloads.10gen.com/win32/mongodb-win32-x86_64-enterprise-windows-64-latest.zip"
+         MONGODB_LATEST="http://downloads.10gen.com/windows/mongodb-windows-x86_64-enterprise-latest.zip"
              MONGODB_42="http://downloads.10gen.com/win32/mongodb-win32-x86_64-enterprise-windows-64-${VERSION_42}.zip"
              MONGODB_40="http://downloads.10gen.com/win32/mongodb-win32-x86_64-enterprise-windows-64-${VERSION_40}.zip"
              MONGODB_36="http://downloads.10gen.com/win32/mongodb-win32-x86_64-enterprise-windows-64-${VERSION_36}.zip"
```
ShaneHarvey: Can you remove the MONGODB_LATEST links for  "cygwin*-i686)" and "windows32*)"? 
https://fastdl.mongodb.org/win32/mongodb-win32-i386-latest.zip is MongoDB 3.3.3

Can you also make all the versions in `windows64*)` match `cygwin*-x86_64)`? I think those are the same platform.

PR #96, opened 2020-03-11
use enterprise links for Ubuntu 18.04 ARM

samantharitter:


samantharitter:
@mbroadst I know!  I only noticed it because FLE wouldn't work on ARM once I upgraded, turns out mongocryptd just wasn't there 🤦‍♀

PR #97, opened 2020-03-12
SPEC-1587 Update Windows stapled OCSP test plan

vincentkam:
https://jira.mongodb.org/browse/SPEC-1587

See also: https://github.com/mongodb/specifications/pull/767


.evergreen/orchestration/configs/servers/windows/rsa-basic-tls-ocsp-mustStaple.json
```diff
@@ -0,0 +1 @@
+../rsa-basic-tls-ocsp-mustStaple.json
```
vincentkam: Symlink, since the files are currently identical

ShaneHarvey: Does this work? I remember symlinks causing problems in windows a few years ago.

vincentkam: Hm a good question! I'd forgotten this runs in Cygwin on the Windows hosts. Symlinks are supposed to work on Windows with Git on Windows 10 in dev mode, but for simplicity's sake, I'll use a copy.

.evergreen/orchestration/configs/servers/windows/README.md
```diff
@@ -2,10 +2,14 @@ These "disableStapling" configuration files have the OCSP failpoint
 removed because the failpoint results in the mongod erroring out on
 Windows.
```
kevinAlbs: Hah I missed that this directory existed. Are these separate configs necessary? Stapling is not supported in macOS either, but I think the failpoint is just a no-op in Windows/macOS based on the comments in the server code:
https://github.com/mongodb/mongo/search?q=disablestapling&unscoped_q=disablestapling
Testing on my Windows machine with  mongod v4.5.0-138-gac9caed I'm able to start it with that failpoint:
`mongod.exe --setParameter enableTestCommands=1 --setParameter failpoint.disableStapling="{'mode': 'alwaysOn'}"`

Additionally, my WIP for windows testing is able to at least start mongod on Windows with the other configs (and fails later for other reasons). Here's one that starts with rsa-basic-tls-ocsp-mustStaple-disableStapling.json: https://evergreen.mongodb.com/task_log_raw/mongo_c_driver_ocsp_ocsp_winssl_malicious_server_test_1_rsa_delegate_patch_d6ced5f6a02fef6be59eba09917f1b788e3d5bec_5e6984d83e8e86487c4dda6f_20_03_12_00_40_23/0?type=T#L412

So perhaps we can remove this directory entirely?

vincentkam: Oh nice! It didn't always use to be a no-op. You're right we can remove this directory!

PR #101, opened 2020-03-16
DRIVERS-932 Resync MONGODB-AWS auth test scripts from the server

bazile-clyde:


bazile-clyde:
@ShaneHarvey I don't see the `RequestLimitExceeded` in the C Driver. Could you verify this solves the issue for Python?

PR #106, opened 2020-04-28
DRIVERS-1216: Add single responder OCSP certificates to drivers-evergreen-tools

vincentkam:
https://jira.mongodb.org/browse/DRIVERS-1216

See https://github.com/vincentkam/mongo/commit/82e70b7420bd0a742c274cbeb920a1282ba5a1a9 for the changes to `mkcert.py`


.evergreen/ocsp/certs.yml
```diff
@@ -49,9 +49,31 @@ certs:
     keyUsage: [digitalSignature, keyEncipherment]
     extendedKeyUsage: [serverAuth, clientAuth]
 
-- name: 'server-mustStaple.pem'
```
divjotarora, line 52: This change isn't clear to me. Aren't drivers relying on a `server-mustStaple.pem` file to exist?

vincentkam: Ah thanks for the catch! I had mutated this definition instead of creating a new one.  I've restored the old definition (the corresponding pem file was never changed)

.evergreen/ocsp/certs.yml
```diff
@@ -87,7 +108,7 @@ certs:
     subjectAltName:
       DNS: localhost
       IP: 127.0.0.1
-    authorityInfoAccess: 'OCSP;URI:http://localhost:9001/power/level,OCSP;URI:http://localhost:8100/status'
+    authorityInfoAccess: 'OCSP;URI:http://localhost:8100/status'
```
vincentkam, line 111: I missed this the first time I generated the singleEndpoint variants, and have regenerated the associated cert as well. I realize now that technically we could probably skip this cert since `mongods` handle multiple endpoints without any issues, but it does no harm to have this cert.  I suppose a driver that does not support stapling but only supports a single endpoint could also use this cert for testing

vincentkam:
Ready for re-review!

vincentkam:
@divjotarora Sure a final sanity check couldn't hurt (as long it isn't too much trouble: if it is, I'm happy that the Files Changed tab shows that no existing files were changed outside of certs.yml)!

divjotarora:
Might take a while, but I kicked off https://evergreen.mongodb.com/version/5ea886102fbabe53640a95a0.

PR #108, opened 2020-05-04
Fix 4.4 download link for linux-amzn-2-x86_64

tfogo:
I think this was a typo? It's causing a build failure for mongomirror testing.


ShaneHarvey:
Yep, looks like a typo to me.

Issue #43, opened 2018-05-24
Failure to install mongo-orchestration on Amazon Linux with Python 2.6


ShaneHarvey:
```
[2018/05/23 10:43:16.398] + pip install --upgrade git+git://github.com/mongodb/mongo-orchestration@master
 [2018/05/23 10:43:16.678] DEPRECATION: Python 2.6 is no longer supported by the Python core team, please upgrade your Python. A future version of pip will drop support for Python 2.6
 [2018/05/23 10:43:16.684] Installing setuptools<37, pip, wheel<0.30...done.
 [2018/05/23 10:43:16.684] Collecting git+git://github.com/mongodb/mongo-orchestration@master
 [2018/05/23 10:43:16.685]   Cloning git://github.com/mongodb/mongo-orchestration (to master) to ./db/pip-XMN79s-build
 [2018/05/23 10:43:17.385] /data/mci/990a07987d5723ebf1ca6d02a5d77ee9/drivers-tools/.evergreen/orchestration/venv/lib/python2.6/site-packages/pip/_vendor/urllib3/util/ssl_.py:339: SNIMissingWarning: An HTTPS request has been made, but the SNI (Subject Name Indication) extension to TLS is not available on this platform. This may cause the server to present an incorrect TLS certificate, which can cause validation failures. You can upgrade to a newer version of Python to solve this. For more information, see https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
 [2018/05/23 10:43:17.385]   SNIMissingWarning
 [2018/05/23 10:43:17.385] /data/mci/990a07987d5723ebf1ca6d02a5d77ee9/drivers-tools/.evergreen/orchestration/venv/lib/python2.6/site-packages/pip/_vendor/urllib3/util/ssl_.py:137: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. You can upgrade to a newer version of Python to solve this. For more information, see https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
 [2018/05/23 10:43:17.385]   InsecurePlatformWarning
[...]

PR #111, opened 2020-05-19
Use Python 3 from the toolchain when available

ShaneHarvey:
Fixes: https://github.com/mongodb-labs/drivers-evergreen-tools/issues/43

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -400,7 +400,8 @@ pre:
   - func: "install dependencies"
 
 post:
-  - func: "upload working dir"
+  # Skip: uploading the full working directory is not needed by drivers-evergreen-tools.
+  # - func: "upload working dir"
   - func: "upload mo artifacts"
   - func: "upload test results"
   - func: "stop mongo-orchestration"
--- a/.evergreen/start-orchestration.sh
+++ b/.evergreen/start-orchestration.sh
@@ -17,6 +17,13 @@ echo From shell `date` > $MONGO_ORCHESTRATION_HOME/server.log
 cd "$MONGO_ORCHESTRATION_HOME"
 # Setup or use the existing virtualenv for mongo-orchestration.
 # Prefer using Python 3 from the toolchain over the default system python.
+# We may still fallback to using an old version of Python here.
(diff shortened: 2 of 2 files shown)


.evergreen/config.yml
```diff
@@ -400,7 +400,8 @@ pre:
   - func: "install dependencies"
 
 post:
-  - func: "upload working dir"
+  # Skip: uploading the full working directory is not needed by drivers-evergreen-tools.
+  # - func: "upload working dir"
```
ShaneHarvey, line 404: I've never needed to download the drivers/working-dir archives. They can be very large (300+ MB on Windows) and can take multiple minutes to upload. I don't think we need this.

.evergreen/start-orchestration.sh
```diff
@@ -16,27 +16,24 @@ echo From shell `date` > $MONGO_ORCHESTRATION_HOME/server.log
...
+# Prefer using Python 3 from the toolchain over the default system python.
+PYTHON=$(command -v /opt/mongodbtoolchain/v2/bin/python3 || command -v python3 || command -v python)
+$PYTHON -c 'import sys; print(sys.version)'
+$PYTHON -c 'import ssl; print(ssl.OPENSSL_VERSION)'
+
 if [ -f venv/bin/activate ]; then
   . venv/bin/activate
 elif [ -f venv/Scripts/activate ]; then
   . venv/Scripts/activate
-elif python -m virtualenv --system-site-packages --never-download venv || virtualenv --system-site-packages --never-download venv; then
```
ShaneHarvey, line 30: I think we should use maintained Python 3 versions where ever possible and avoid using `--never-download`. @behackett do you see a problem with this new approach?

behackett: You're assuming the python3 in the server toolchain is still supported by the latest setuptools and pip, or whatever "python3" points to on the OS, or whatever "python" points to on the OS. I don't think you can make that assumption.

ShaneHarvey: Okay I added back `--never-download`. It's worth mentioning that `--never-download` defaults to True in recent virtualenv versions. I did not track down the specific version it was changed. But regardless I agree that it will help.

PR #115, opened 2020-07-02
SPEC-1684: Testing Atlas Data Lake for Drivers

jstewart-mongo:
Related to https://github.com/mongodb/specifications/pull/821


.evergreen/run-mongohouse-local.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/sh
+#
+# This script builds and launches a local mongohoused for testing.
```
jmikola, line 3: Is there a particular reason this isn't co-located in the `atlas_data_lake` directory? I think that would be consistent with what is done for some of the other projects included in the shared repository, and help keep the common path from becoming too cluttered.

Additionally, a basic README for the `atlas_data_lake` directory would be helpful. At the very least, it can cross-reference the tentative path that will be merged to the specifications repo in https://github.com/mongodb/specifications/pull/821 and briefly explain how this script would be integrated into a driver's Evergreen config.

See [`.evergreen/ocsp`](https://github.com/mongodb-labs/drivers-evergreen-tools/tree/master/.evergreen/ocsp) for some prior art. I realize that README doesn't do everything I asked for above, but it's still helpful in providing some context beyond just having a shell script and config file.

.evergreen/atlas_data_lake/testdata/stores/drivers-tests/test/1.json
```diff
@@ -0,0 +1,15 @@
+{"a": 1, "b": 2, "c": 3}
```
jmikola, line 1: Does the `test/1.json` correspond to the namespace? If so, I imagine this file is defining data for a collection named "1" in the "test" database, which would conflict with the assertions in the https://github.com/mongodb/specifications/pull/821 integration tests.

If so, please rename accordingly.

.evergreen/run-mongohouse-local.sh
```diff
@@ -0,0 +1,67 @@
...
+# There is no corresponding 'shutdown' script; this project relies
+# on Evergreen to terminate processes and clean up when tasks end.
+
+set -o xtrace   # Write all commands first to stderr
+set -o errexit  # Exit the script with error if any of the commands fail
+
+ORIG_DIR="$(pwd)"
+
+# Configure git to use the git protocol
+git config --global url."git@github.com:".insteadof "https://github.com/"
```
jmikola, line 14: I didn't see this done in any other files in the repository. Is this necessary, given that `git clone` below is using the SSH address?

`git config --global` gave me pause here, since that could potentially change the system configuration for anything that executes after this script.

.evergreen/atlas_data_lake/drivers-test-local.yaml
```diff
@@ -0,0 +1,66 @@
...
+            collections:
+              - name: driverdata
+                dataSources:
+                  - storeName: localstore
+                    path: /test/*
+  sqlSchema:
+    inline:
+      - database: "test"
+        collection: "driverdata"
+        schemaFile: "${DRIVERS_TOOLS}/.evergreen/atlas_data_lake/schema/testdata/driver-data.json"
```
jmikola, line 66: In the interest of simplicity, can we refer to `.evergreen/atlas_data_lake/schema.json` or `.evergreen/atlas_data_lake/schema/test.driverdata.json` instead? I don't think "testdata" appears anywhere else, so I assume that doesn't provide much context. Also, I think it'd be clearer to use the namespace in the schema filename.

Likewise, can `drivers-test-local.yaml` be renamed to `config.yaml`. That would be self-documenting by immediately looking at the directory, and I don't imagine we'd need multiple configuration files for this project.

.evergreen/atlas_data_lake/drivers-test-local.yaml
```diff
@@ -0,0 +1,66 @@
...
+                salt: "6i+wBCHE21EaI8IF9+tvPkCoYWY+IyedLaVXGQ=="
+                storedKey: "hnn0iJZ6ZNO9X34oyxtZUKLAMScwpCdch3vGPuDXOvM="
+                serverKey: "3Qg6ptTiTvPwhGSZi0NpAkmsW8P5C/J46Is1t7xaIho="
+            roles:
+              - role: atlasAdmin
+      storage:
+        stores:
+          - name: localstore
+            provider: local
+            path: "${DRIVERS_TOOLS}/.evergreen/atlas_data_lake/testdata/stores/drivers-tests"
```
jmikola, line 54: If I understand correctly, `stores` need only refer to a single directory. Can we just use `.evergreen/atlas_data_lake/stores` instead and then use `test.driverdata.json` as the path below? That might let us eliminate the `/test/*` wildcard and cryptic `1.json` filename.

I understand why mongohouse supports wildcards to populate a collection with data from many sources, but I don't think we actually need to utilize that in our testing (and aren't, if we're just using a single JSON file anyway), so I'd prefer to have a simpler directory structure in the interest of future maintainers.

.evergreen/atlas_data_lake/drivers-test-local.yaml
```diff
@@ -0,0 +1,66 @@
...
+tenants:
+  config:
+    inline:
+      hostnames:
+        - localhost
+      security:
+        roles:
+          # Based on the atlasAdmin custom role defined in NDSDBUser.java in the
+          # 10gen/mms repo.
+          - role: atlasAdmin
```
jmikola, line 23: Do we actually need all of this for the driver ADL testing? I would think `readAnyDatabase` would be sufficient, with no additional privileges.

My concern here is just paring this down to the bare minimum required to test and eliminating anything that was casually copied from another source -- in the interest of making this easier to maintain/understand down the line.

.evergreen/atlas_data_lake/drivers-test-local.yaml
```diff
@@ -0,0 +1,66 @@
+environment: development
```
jmikola, line 1: Can this file use `.yml` as its extension for consistency with https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/ocsp/certs.yml and most of our spec tests?

.evergreen/atlas_data_lake/drivers-test-local.yaml
craiggwilson: This role always exists and is sent to ADL from Atlas (when Atlas is used as the control plane). In this case, it is mimicking what Atlas gives us.

But no, it probably doesn't need to exist if the `mhuser` role below is assigned a different role, like `readAnyDatabase`.

.evergreen/atlas_data_lake/drivers-test-local.yaml
craiggwilson: The `sqlSchema` section can be removed completely if we are _not_ using the `$sql` stage in an aggregation pipeline.

.evergreen/run-mongohouse-local.sh
craiggwilson: This is necessary because one of the dependencies, `mongoast`, is a private repository and the go tool, during build, can't resolve a private repository without this bit.

.evergreen/atlas_data_lake/drivers-test-local.yaml
craiggwilson: no restriction here from ADL.

.evergreen/run-mongohouse-local.sh
jmikola: Thanks for context. @jstewart-mongo: can you summarize the above justification into a comment for the benefit of future maintainers that come across this file?

.evergreen/atlas_data_lake/drivers-test-local.yaml
jmikola: None of the `aggregation` tests in https://github.com/mongodb/specifications/pull/821 use `$sql`, so I believe this can probably be removed entirely.

.evergreen/atlas_data_lake/drivers-test-local.yaml
jmikola: Thanks for clarifying. I'd opt for just using `readAnyDatabase` if that's all we need to run the tests in https://github.com/mongodb/specifications/pull/821 then.

.evergreen/atlas_data_lake/config.yml
```diff
@@ -46,21 +31,16 @@ tenants:
...
               - name: driverdata
                 dataSources:
                   - storeName: localstore
-                    path: /test/*
-  sqlSchema:
-    inline:
-      - database: "test"
-        collection: "driverdata"
-        schemaFile: "${DRIVERS_TOOLS}/.evergreen/atlas_data_lake/schema/testdata/driver-data.json"
+                    path: test.driverdata.json
```
jmikola, line 46: Ah, this is even simpler. :+1:

.evergreen/atlas_data_lake/README.md
```diff
@@ -0,0 +1,9 @@
+
```
jmikola, line 1: Suggested title: `# Atlas Data Lake Testing`

jstewart-mongo: I have added the title.

jyemin:
Removed myself as a reviewer

jmikola:
For the benefit of anyone else following this PR, this was merged in https://github.com/mongodb-labs/drivers-evergreen-tools/commit/abe66fdc0d1c58d92a97e17d4218baca884ee76d.

PR #117, opened 2020-07-15
Define admin role for ADL testing

jmikola:
This adds an admin role based on the mongohouse config examples, which defines additional permissions for extra operations that may be performed by the driver's test runner (e.g. dropping collections, querying the server configuration).


jmikola:
@jyemin: I believe we both confirmed that this is no longer necessary once we disable behavior in the spec test runner to drop the collection under test before executing operations (see: https://github.com/mongodb/specifications/pull/839). Closing without a merge.

PR #119, opened 2020-07-21
Split mongohoused into separate build/run scripts

divjotarora:
The Java Driver commit for ADL testing runs the `bootstrap-mongohoused` Evergreen task in the background (https://github.com/mongodb/mongo-java-driver/commit/7f6e0c3be351e9d586ca4a6271f79af654b03deb#diff-fdcefd1312a87e7c09196afa0f0633a3R227) because otherwise the server would run in the foreground and block progress for the rest of the tasks. When adding this to the Go Driver, I found that the build was taking too long and the Go tests started running before the build was complete, so they failed because there was no server running.

This PR splits the existing `run-mongohouse-local.sh` into separate `build-mongohouse-local.sh` and `run-mongohouse-local.sh` scripts. These first would be called in the foreground and the second would be in the background. The actual server startup is very quick so the drivers tests should discover the server within server selection timeout even if they start running before the server has started. See https://github.com/divjotarora/mongo-go-driver/commit/39969242bbb9d5c1c788884025a231c77bb4e5dd#diff-c291812f12d2ddfd4045a0c39bf9d130R165 for an example invocation of these scripts in the Go Driver.

### DIFF
--- a/.evergreen/atlas_data_lake/README.md
+++ b/.evergreen/atlas_data_lake/README.md
@@ -2,7 +2,13 @@
 
 The files in this directory are required to build and launch a local
 `mongohoused` for testing. In the Evergreen configuration file, execute
-the following command:
+the following command in the foreground to build the server:
+
+```
+sh ${DRIVERS_TOOLS}/.evergreen/atlas_data_lake/build-mongohouse-local.sh
+```
+
+and the following command in the background to run the server:
 
 ```
 sh ${DRIVERS_TOOLS}/.evergreen/atlas_data_lake/run-mongohouse-local.sh
--- a/.evergreen/atlas_data_lake/build-mongohouse-local.sh
+++ b/.evergreen/atlas_data_lake/build-mongohouse-local.sh
@@ -1,0 +1,63 @@
+#!/bin/sh
+#
+# This script builds a local mongohoused for testing.
+
+set -o xtrace   # Write all commands first to stderr
+set -o errexit  # Exit the script with error if any of the commands fail
+
+ORIG_DIR="$(pwd)"
+
+# Configure git to use the git protocol
+git config --global url."git@github.com:".insteadof "https://github.com/"
+
+AP_START=$(date +%s)
+
+# Set up environment variables for Go
+GO_VERSION=1.14
+export PATH="/opt/golang/go${GO_VERSION}/bin:$PATH"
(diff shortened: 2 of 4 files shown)


jyemin:
How long was it taking for you?  I'm guessing it's working in the Java driver because the mongohoused build completes in less than the default server selection timeout of 30 seconds.

divjotarora:
It took longer than 30 seconds. Also, our testing framework tries to run `ping` with a 2 second timeout so we can fail fast if there's no server running. I'd prefer not to change that long-term and changing it to 30 seconds still caused the patch build to fail. I ran the `run-mongohouse-local.sh` script on a spawnhost and saw it consistently take more than 30 seconds as well.

jyemin:
Yep, i checked the Java build and the reason it probably works is because it takes several minutes before the first test is actually ready to run.  Which gives mongohoused more than enough time to build in the background.

divjotarora:
Ran locally and the whole script took ~2 minutes. The Go tests are ready to run much earlier because the driver repository has already been cloned and the build time is less than 5 seconds.

mbroadst:
should we be using binary builds of mongohoused rather than building it every time a driver has a CI run?

jmikola:
> should we be using binary builds of mongohoused rather than building it every time a driver has a CI run?

What are our options for doing so? Does Evergreen support a cache (like [Travis](https://docs.travis-ci.com/user/caching/)) where we could just ask it to cache the build directory?

Alternatively, if mongohouse published [releases](https://github.com/10gen/mongohouse/releases) I suppose we could grab them from there -- although we'd still need to provision `mqlrun` (not sure where that comes from).

Your question did make me consider that the current CI config is always testing the `master` branch of mongohoused. Surely we'd at least want to standardize on a stable release, no? Unfortunately, the scope for [DRIVERS-829](https://jira.mongodb.org/browse/DRIVERS-829) never discussed exactly what version of ADL we want to test so I think this was entirely overlooked.

divjotarora:
@jmikola Updated the README.

divjotarora:
@jmikola @mbroadst I agree that testing against released versions or otherwise pre-built binaries would be better, but it seems out of scope for this PR. I've filed https://jira.mongodb.org/browse/DRIVERS-1341 to address that concern.

divjotarora:
Note that this PR now includes a change to the ADL config file to account for https://jira.mongodb.org/browse/MHOUSE-1556.

PR #121, opened 2020-09-01
Add --bind_ip argument to ocsp mock

p-mongo:
This is needed when running the mock in Docker
and using it from the host system.


ShaneHarvey:
@p-mongo, can you open a server ticket to get this change merged upstream? That way we don't need to apply this patch every time we re-sync this file from the server.

p-mongo:
https://jira.mongodb.org/browse/SERVER-50690 @ShaneHarvey

PR #125, opened 2020-10-20
PYTHON-2298 Reset macOS file descriptor ulimit to 64000 to fix mongod errors

ShaneHarvey:
Somewhat frequently we run on a macOS host that has a ulimit -n set to 256 which causes mongod to crash with `Too many open files `. This change resets the limit to the usual 6400.

Here's an example patch that shows the error behavior. Note: when mongod crashes the python test suite times out after 30min so we're only interested in failures due to timeouts (we ignore the other task failures): https://evergreen.mongodb.com/version/5f8a1c1cd6d80a7b025c03cb

An here's a patch that tests the new behavior:
https://evergreen.mongodb.com/version/5f8f4c740ae60616c87a1646

### DIFF
--- a/.evergreen/start-orchestration.sh
+++ b/.evergreen/start-orchestration.sh
@@ -67,8 +67,6 @@ fi
 # TooManyFilesOpen: 24: Too many open files
 if [ "$(uname -s)" = "Darwin" ]; then
   ulimit -a
-  ulimit -n 262144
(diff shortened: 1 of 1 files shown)


ShaneHarvey:
Closing, the evergreen team is working on a fix: https://jira.mongodb.org/browse/EVG-13244

PR #126, opened 2020-11-16
Add ability to set background parameter for tasks

samantharitter:
This change is necessary to support the C driver's new mongohouse test.


kevinAlbs (approved):
LGTM

PR #127, opened 2020-12-07
Support requireApiVersion parameter for standalone servers

alcaeus:
This introduces a new `REQUIRE_API_VERSION` variable into the setup process for mongo-orchestration. For now, the run-orchestration.sh script will connect to the started cluster and run a `setParameter` command to enable `requireApiVersion` if the variable is set to a non-zero value.

Once PyMongo and mongo-orchestration are able to deal with the versioned API we can remove the workaround and start clusters from config files that include the requireApiVersion parameter.

### DIFF
--- a/.evergreen/run-orchestration.sh
+++ b/.evergreen/run-orchestration.sh
@@ -7,6 +7,9 @@ AUTH=${AUTH:-noauth}
 SSL=${SSL:-nossl}
 TOPOLOGY=${TOPOLOGY:-server}
 STORAGE_ENGINE=${STORAGE_ENGINE}
+# Set to a non-empty string to set the requireApiVersion parameter
+# This is currently only supported for standalone servers
+REQUIRE_API_VERSION=${REQUIRE_API_VERSION}
 # Set to a non-empty string to use the <topology>/disableTestCommands.json
 # cluster config, eg DISABLE_TEST_COMMANDS=1
 DISABLE_TEST_COMMANDS=${DISABLE_TEST_COMMANDS}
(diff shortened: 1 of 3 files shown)


.evergreen/run-orchestration.sh
```diff
@@ -95,3 +98,8 @@ cat <<EOT >> $DRIVERS_TOOLS/results.json
 ]}
 
 EOT
+
+# Set the requireApiVersion parameter
+if [ ! -z "$REQUIRE_API_VERSION" ]; then
+  mongo $URI $MONGO_ORCHESTRATION_HOME/require-api-version.js
```
jyemin, line 104: If this is accidentally done against a server that doesn't support require api version, will it cause m-o to fail (hoping yes)

alcaeus: I was hoping it would, but it looks like `mongo` returns a 0 exit code even when the command fails:
```
$ mongo .evergreen/orchestration/require-api-version.js 
MongoDB shell version v4.2.11
connecting to: mongodb://127.0.0.1:27017/?compressors=disabled&gssapiServiceName=mongodb
Implicit session: session { "id" : UUID("8610a0d0-8064-4066-b2d5-29601a33025a") }
MongoDB server version: 4.2.11

$ echo $?
0
```

However, this also holds true for using `eval` when the command fails:

```
$ mongo --eval 'db.adminCommand({ "setParameter": 1, "requireApiVersion": 1 });'
MongoDB shell version v4.2.11
connecting to: mongodb://127.0.0.1:27017/?compressors=disabled&gssapiServiceName=mongodb
Implicit session: session { "id" : UUID("c43b0af6-0c0e-4ac9-96ae-7c1d429d6992") }
MongoDB server version: 4.2.11
{
	"ok" : 0,
	"errmsg" : "attempted to set unrecognized parameter [requireApiVersion], use help:true to see options "
}

$ echo $?
0
```

The only way I could come up with is to combine `--eval` with parsing the response to make sure the command succeeded, but I don't think this is worth it.

If there are no objections, I'd merge this down hoping that we can have versioned API support in m-o sooner than later.

jyemin (approved):
LGTM

PR #129, opened 2020-12-23
Add a mongo-orchestration config for versioned API tests

alcaeus:
The new mongo-orchestration config starts a standalone server that enables the `acceptAPIVersion2` parameter. This is required to run the versioned API spec tests that test deprecated commands.

This config is only compatible with server versions 4.9 and newer and will fail to start on older server versions. Drivers can use this config in place of the `basic` config when testing against `latest`, but may also choose to create a separate build config that only tests versioned API tests against this config specifically.

// cc @benjirewis since I can't add you as a reviewer.

### DIFF
--- a/.evergreen/orchestration/configs/servers/versioned-api-testing.json
+++ b/.evergreen/orchestration/configs/servers/versioned-api-testing.json
@@ -1,0 +1,12 @@
+{
+  "id": "versioned-api-testing",
+  "name": "mongod",
+  "procParams": {
+    "ipv6": true,
+    "bind_ip": "127.0.0.1,::1",
+    "logappend": true,
+    "journal": true,
+    "port": 27017,
+    "setParameter": {"enableTestCommands": 1, "acceptAPIVersion2":  1}
+  }
+}


jyemin:
Can you post a link to a patch build of a driver that uses this new config?

alcaeus:
https://spruce.mongodb.com/version/5fe346ed3e8e862d05f92c2e/tasks tests this in the rust driver. Note that the SSL builds fail because I didn't bother to set this up properly, the non-tls tests work. I'll wrap up the rust tests after this PR has been merged.

benjirewis:
LGTM, thanks @alcaeus 🕵️‍♂️  

As for the failing SSL builds though, I think tests that set `REQUIRE_API_VERSION` will always fail on SSL because [require-api-version.js](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/d880186c76e1b61f55bd42a18078fd0fb04fec6b/.evergreen/orchestration/require-api-version.js) sends a command to the server without SSL. If we're thinking about testing with SSL, I think a small change will be required for `require-api-version.js` as well.

alcaeus:
The problem with `require-api-version.js` is known - that itself is a stop-gap solution until PyMongo and mongo-orchestration support the versioned API, after which we'll make sure to add configurations that set the `requireApiVersion` server parameter through the orchestration configs instead of enabling it after the fact.

ShaneHarvey:
For server parameters like this which I assume can be changed at runtime, would it be easier to have the driver's test suite run `setParameter` itself rather than involve all the ceremony of drivers-evergreen-tools and mongo-orchestration?

jyemin:
@ShaneHarvey pretty sure that this one can not be changed at runtime, hence the ceremony.

ShaneHarvey:
Gotcha, should we ask the server to allow it to be changed at runtime? That would make this testing easier.

alcaeus:
Revisiting this as @benjirewis reminded me this was still pending. To explain the unusual setup here:
* `acceptAPIVersion2` cannot be changed at runtime, so we have to set this in the mo config. I don't think it's necessary to change the server to allow changes at runtime.
* `requireApiVersion` can be changed at runtime, but the only reason we use that method of enabling it is that until PyMongo and mongo-orchestration support the versioned API, we can't set this in the config.

This config is relevant for running the versioned API spec tests, which don't need to be run (and in some cases can't be run) on a server with `requireApiVersion` enabled. I can create a separate PR to ensure that the `requireApiVersion` workaround currently necessary also works on SSL, but I'd rather not do this in this PR to unblock testing for Go.

PR #25, opened 2017-11-07
Start shards as three member replica sets.

ShaneHarvey:
Resolves #24.

Evergreen: https://evergreen.mongodb.com/version/5a020a83e3c3316365000399

PR #133, opened 2021-01-20
Use three-member replica sets for shards

alcaeus:
After #25, this is the second attempt to introduce three-member replica sets for shards. Note that I've made this a PSS replica set, while the original change used a PSA replica set. Let me know if I should change this.

### DIFF
--- a/.evergreen/orchestration/configs/sharded_clusters/auth-ssl.json
+++ b/.evergreen/orchestration/configs/sharded_clusters/auth-ssl.json
@@ -7,14 +7,32 @@
     {
       "id": "sh01",
       "shardParams": {
-        "members": [{
-          "procParams": {
-            "ipv6": true,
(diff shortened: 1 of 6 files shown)


jyemin:
Remind me, what's driving this change?

alcaeus:
PHP currently overrides a lot of scripts from this repository as it runs some tests with secondary read preferences that fail on sharded clusters where no secondary exists. I figured it would make sense to use three members, especially considering that's the setup users would typically use.

ShaneHarvey (approved):
LGTM, please notify the rest of the drivers team so we all know to look out for potential sharded cluster test failures once this is merged.

ShaneHarvey:
Testing this change with the python driver here: https://evergreen.mongodb.com/version/6008b1342a60ed0a440b157f

jyemin (approved):

PR #136, opened 2021-01-22
SPEC-1774 add CSFLE temp credentials script

kevinAlbs:
Includes scripts to obtain temporary AWS credentials for testing CSFLE.

### DIFF
--- a/.evergreen/csfle/set-temp-creds.sh
+++ b/.evergreen/csfle/set-temp-creds.sh
@@ -22,8 +22,18 @@
 
 set +o xtrace # Disable tracing.
 
+get_creds() {
+    $PYTHON - "$@" << 'EOF'
+import boto3
+
+client = boto3.client('sts')
+credentials = client.get_session_token()["Credentials"]
+print (credentials["AccessKeyId"] + " " + credentials["SecretAccessKey"] + " " + credentials["SessionToken"])
+EOF
+}
+
 PYTHON=${PYTHON:-python}
-CREDS=$($PYTHON print-temp-creds.py)
+CREDS=$(get_creds)
 
 export CSFLE_AWS_TEMP_ACCESS_KEY_ID=$(echo $CREDS | awk '{print $1}')
 export CSFLE_AWS_TEMP_SECRET_ACCESS_KEY=$(echo $CREDS | awk '{print $2}')
--- a/.evergreen/csfle/set-temp-creds.sh
+++ b/.evergreen/csfle/set-temp-creds.sh
@@ -1,0 +1,30 @@
+# Obtains temporary AWS credentials for CSFLE testing.
+#
+# Run with a . to add environment variables to the current shell:
+# . ./set-temp-creds.sh
+#
+# Requires the python AWS SDK boto3. This can be installed with: pip install boto3
+# The path to python in a virtual environment may be passed with the PYTHON
+# environment variable.
+#
+# Requires AWS credentials for CSFLE to obtain temporary credentials.
+# Those credentials can be passed through the following environment variables:
+#
+# export AWS_ACCESS_KEY_ID=...
+# export AWS_SECRET_ACCESS_KEY=...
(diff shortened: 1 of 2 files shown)


rozza (approved):

.evergreen/csfle/set-temp-creds.sh
```diff
@@ -0,0 +1,30 @@
...
+# . ./set-temp-creds.sh
+#
+# Requires the python AWS SDK boto3. This can be installed with: pip install boto3
+# The path to python in a virtual environment may be passed with the PYTHON
+# environment variable.
+#
+# Requires AWS credentials for CSFLE to obtain temporary credentials.
+# Those credentials can be passed through the following environment variables:
+#
+# export AWS_ACCESS_KEY_ID=...
```
saghm, line 13: Is there a reason we don't want to just export the credentials with the env var names that actually are used by the driver?

.evergreen/csfle/set-temp-creds.sh
```diff
@@ -0,0 +1,30 @@
...
+# After running this script, the following shell variables are set:
+# - CSFLE_AWS_TEMP_ACCESS_KEY_ID
+# - CSFLE_AWS_TEMP_SECRET_ACCESS_KEY
+# - CSFLE_AWS_TEMP_SESSION_TOKEN
+#
+
+set +o xtrace # Disable tracing.
+
+PYTHON=${PYTHON:-python}
+CREDS=$($PYTHON print-temp-creds.py)
```
saghm, line 26: Rather than relying on this script to get invoked in the same directory as the Python script, maybe we could use a heredoc? (see https://unix.stackexchange.com/questions/624875/is-there-a-way-to-use-heredoc-for-bash-and-zsh-and-be-able-to-use-arguments#answer-624892)

```
get_creds() {
    $PYTHON - "$@" << 'EOF'
import boto3

client = boto3.client('sts')
credentials = client.get_session_token()["Credentials"]
print (credentials["AccessKeyId"] + " " + credentials["SecretAccessKey"] + " " + credentials["SessionToken"])
EOF
}

PYTHON=${PYTHON:-python}
CREDS=$(get_creds)
```

It's messy, but it will make it easier for people to use, and honestly, when is bash _not_ messy?

.evergreen/csfle/set-temp-creds.sh
kevinAlbs: Perhaps I am misunderstanding the question. CSFLE credentials cannot be configured with environment variables, so the output variable names I chose are arbitrary. The input variable names are consistent with MONGODB-AWS auth (with the addition of AWS_DEFAULT_REGION) to pass along to boto3.

.evergreen/csfle/set-temp-creds.sh
kevinAlbs: Oh thank you for writing that out. One less script does seem like a win.

> when is bash not messy?

Good point!

.evergreen/csfle/set-temp-creds.sh
saghm: The comment here says that these credentials need to be passed through to `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, etc. to be used. I'm wondering why we don't just set those variables directly from within this script rather than requiring users to manually set them from the variables that we set instead in this script.

saghm: Based on offline discussion, I misunderstood what this comment was saying and thought that the variables set by this script needed to be set to `AWS_ACCESS_KEY_ID` and the others mentioned. Given that this is not the case, I don't think any changes need to be made from where the script was before my previous comment; re-exporting the variables from within the script only will do something if they're already set, which means nothing will actually be changed in that case.

kevinAlbs: Reverted!

PR #138, opened 2021-01-28
DRIVERS-1462: Add script for using venv for AWS auth

emadum:
Adds a utility script that can be used to reduce boilerplate around using virtualenv for the AWS auth tests.

### DIFF
--- a/.evergreen/auth_aws/activate_venv.sh
+++ b/.evergreen/auth_aws/activate_venv.sh
@@ -1,7 +1,7 @@
 set +x
 
 if [ "Windows_NT" = "$OS" ]; then
-  PYTHON_BINARY=python.exe
+  PYTHON_BINARY=C:/python/Python38/python.exe
 else
   PYTHON_BINARY=python3
 fi
--- a/.evergreen/auth_aws/activate_venv.sh
+++ b/.evergreen/auth_aws/activate_venv.sh
@@ -1,0 +1,25 @@
+set +x
+
+if [ "Windows_NT" = "$OS" ]; then
+  PYTHON_BINARY=python.exe
+else
+  PYTHON_BINARY=python3
+fi
+
+# create venv on first run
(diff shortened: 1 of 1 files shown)


.evergreen/auth_aws/activate_venv.sh
```diff
@@ -0,0 +1,25 @@
+set +x
+
+if [ "Windows_NT" = "$OS" ]; then
+  PYTHON_BINARY=python.exe
```
ShaneHarvey, line 4: In pymongo and .NET on windows we use:
```
  PYTHON_BINARY=C:/python/Python38/python.exe
```

ShaneHarvey (commented):
Neat!

.evergreen/auth_aws/activate_venv.sh
emadum: More explicit is probably better, but I wanted to use the same logic as [here](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/auth_aws/lib/aws_e2e_lib.js#L33-L39) to make sure they are using the same python; I wonder if it'd make sense to update that in both places?

ShaneHarvey: The logic in that file assumes the javascript file is running inside a virtualenv. The logic here to create that virtualenv is different. We have to change this to:
```
PYTHON_BINARY=C:/python/Python38/python.exe
```
Or else we'll end up using Python 2.7 which won't work.

>  I wonder if it'd make sense to update that in both places?

Nope the js file should remain the same. It's a quirk that windows virtualenvs don't create an alias for "python3", only  "python".

ShaneHarvey (approved):
LGTM! Tests pass with python on macOS, Ubuntu, and Windows here: https://evergreen.mongodb.com/version/6013259032f41771a81668fe

PR #139, opened 2021-02-03
PYTHON-2537 Fix mongodb download links for Centos 6

ShaneHarvey:
This change fixes the download links for Centos 6 which we use for performance testing. Centos 6 is compatible with the RHEL 6.2 releases. Tested here: https://spruce.mongodb.com/version/601aeb54d1fe0705740e9e26

The old behavior was that Centos 6 used the "generic" linux releases which only support up to MongoDB 4.0. Now all versions are supported (including TLS support).


behackett (approved):
Neat!

PR #141, opened 2021-03-25
DRIVERS-1596 Remove usages of xtrace in shell scripts

alcaeus:
DRIVERS-1596

The `xtrace` shell option prints all commands that are executed, which may reveal sensitive information. It also clutters the build logs, making debugging issues a lot harder.

While it's useful to use `xtrace` to ensure commands and options are assembled correctly, this should only be used while developing and not as part of the regular usage.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -69,7 +69,7 @@ functions:
            PROJECT_DIRECTORY: "$PROJECT_DIRECTORY"
            PREPARE_SHELL: |
               set -o errexit
-              set -o xtrace
+              set +o xtrace
               export DRIVERS_TOOLS="$DRIVERS_TOOLS"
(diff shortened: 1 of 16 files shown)


.evergreen/config.yml
```diff
@@ -69,7 +69,7 @@ functions:
            PROJECT_DIRECTORY: "$PROJECT_DIRECTORY"
            PREPARE_SHELL: |
               set -o errexit
-              set -o xtrace
+              set +o xtrace
```
divjotarora, line 72: Why the change to `+o xtrace` here rather than removing the existing line as was done in other files?

alcaeus: I had originally left this in to disable xtrace when preparing a shell, but any new shell most likely starts out with it disabled anyways. I've removed it 👍

divjotarora (approved):

PR #142, opened 2021-04-01
DRIVERS-1603 Introduce scripts for provisioning and reaping serverless instances

patrickfreed:
[DRIVERS-1603](https://jira.mongodb.org/browse/DRIVERS-1603)

This PR introduces the scripts necessary for drivers to create and delete serverless instances on demand as part of their test runs, as well as an example evergreen config that uses does so. See [here](https://evergreen.mongodb.com/version/6066352d2a60ed028ace6af7) for an example of the Swift driver using them to run the CRUD tests against a serverless instance.

Once the proxy implements some of the stuff we requested, I'll open up a new PR on the specs repo with drivers instructions on how to use these and what tests to run.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -835,6 +835,9 @@ axes:
       - id: debian92-test
         display_name: "Debian 9.2"
         run_on: debian92-test
+      - id: ubuntu-18.04
+        display_name: "Ubuntu 18.04"
+        run_on: ubuntu1804-test
 
   - id: topology
     display_name: Topology
@@ -1130,9 +1133,9 @@ buildvariants:
 
 - matrix_name: "serverless"
   matrix_spec:
-    os-requires-32:
-      - "ubuntu-16.04"
-  display_name: "Serverless ${os-fully-featured}"
+    os-requires-40:
+      - "ubuntu-18.04"
+  display_name: "Serverless ${os-requires-40}"
   tasks:
     - "serverless_task_group"
 
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1130,8 +1130,8 @@ buildvariants:
 
 - matrix_name: "serverless"
   matrix_spec:
-    os-fully-featured:
-      - "ubuntu-18.04"
+    os-requires-32:
+      - "ubuntu-16.04"
   display_name: "Serverless ${os-fully-featured}"
   tasks:
     - "serverless_task_group"
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -339,6 +339,21 @@ functions:
           ${PREPARE_SHELL}
           AUTH=${AUTH} SSL=${SSL} MONGODB_URI="${MONGODB_URI}" sh ${PROJECT_DIRECTORY}/.evergreen/run-tests.sh
 
+  "run serverless tests":
+    - command: shell.exec
+      type: test
+      params:
+        working_dir: "src"
+        script: |
+          ${PREPARE_SHELL}
+          set +o xtrace
+
+          MONGODB_URI="${MONGODB_URI}" \
+          MONGODB_API_VERSION=${MONGODB_API_VERSION} \
+          SERVERLESS_ATLAS_USER="${SERVERLESS_ATLAS_USER}" \
+          SERVERLESS_ATLAS_PASSWORD="${SERVERLESS_ATLAS_PASSWORD}" \
+            sh ${PROJECT_DIRECTORY}/.evergreen/run-serverless-tests.sh
+
   "cleanup":
     - command: shell.exec
       params:
@@ -679,6 +694,11 @@ tasks:
             TOPOLOGY: "sharded_cluster"
         - func: "run tests"
 
+    - name: "test-serverless"
+      tags: ["serverless"]
+      commands:
+        - func: "run serverless tests"
+
 # }}}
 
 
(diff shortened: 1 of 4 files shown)


.evergreen/config.yml
```diff
@@ -339,6 +339,21 @@ functions:
...
+      params:
+        working_dir: "src"
+        script: |
+          ${PREPARE_SHELL}
+          set +o xtrace
+
+          MONGODB_URI="${MONGODB_URI}" \
+          MONGODB_API_VERSION=${MONGODB_API_VERSION} \
+          SERVERLESS_ATLAS_USER="${SERVERLESS_ATLAS_USER}" \
+          SERVERLESS_ATLAS_PASSWORD="${SERVERLESS_ATLAS_PASSWORD}" \
```
patrickfreed, line 354: Since the URIs for the resulting instances can't be known ahead of time, the username / password need to be supplied separately, and the test runners need to make sure they get passed along. These would need to be supplied as private evergreen variables.

.evergreen/config.yml
```diff
@@ -869,6 +889,39 @@ axes:
         variables:
            STORAGE_ENGINE: "inmemory"
 
+task_groups:
```
patrickfreed, line 892: It's important that drivers use task groups or `post` to ensure the serverless instances get reaped properly, even on test failure.

.evergreen/config.yml
```diff
@@ -1075,6 +1128,14 @@ buildvariants:
            - name: "test-2.4-sharded_cluster"
            - name: "test-2.4-standalone"
 
+- matrix_name: "serverless"
+  matrix_spec:
+    os-fully-featured:
+      - "ubuntu-18.04"
```
patrickfreed, line 1134: I wasn't sure what the platform requirement would be for drivers, so I just included a single example.

.evergreen/serverless/create-instance.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
+
+set -o errexit
+set +o xtrace # disable xtrace to ensure credentials aren't leaked
+
+if [ -z "$PROJECT" ]; then
+    echo "Project name must be provided via PROJECT environment variable"
+    exit 1
+fi
+INSTANCE_NAME="$RANDOM-$PROJECT"
```
patrickfreed, line 10: This is a pretty simple / crude way to ensure the instances have unique names, but it's effective and works within the relatively strict limitations that are set on cluster names: https://docs.atlas.mongodb.com/reference/atlas-limits/#label-limits

.evergreen/config.yml
```diff
@@ -1075,6 +1128,14 @@ buildvariants:
            - name: "test-2.4-sharded_cluster"
            - name: "test-2.4-standalone"
 
+- matrix_name: "serverless"
+  matrix_spec:
+    os-requires-32:
+      - "ubuntu-16.04"
```
kevinAlbs, line 1134: Should this be Ubuntu 18.04? Latest server builds are no longer published on Ubuntu 16.04 (see CDRIVER-3946). It won't matter for serverless. And drivers don't need to copy this. But if drivers are testing as many server variants as possible on as few platforms as they can, 18.04 could be preferable.

.evergreen/config.yml
```diff
@@ -1075,6 +1128,14 @@ buildvariants:
...
+  display_name: "Serverless ${os-fully-featured}"
```
kevinAlbs, line 1135: If I am understanding the matrix definition, I think this should be:
```suggestion
  display_name: "Serverless ${os-requires-32}"
```

.evergreen/config.yml
kevinAlbs: Is this the user/password for the Atlas account? Or is this the user/password to the new cluster? Will all drivers be using the same user/password?

.evergreen/serverless/create-instance.sh
kevinAlbs: Oh clever, TIL about `$RANDOM` (and `$SECONDS`).

.evergreen/serverless/create-instance.sh
```diff
@@ -0,0 +1,82 @@
...
+
+echo "creating new serverless instance \"${INSTANCE_NAME}\"..."
+
+DIR=$(dirname $0)
+API_BASE_URL="https://account-dev.mongodb.com/api/atlas/v1.0/groups/$SERVERLESS_DRIVERS_GROUP"
+
+curl \
+  -u "$SERVERLESS_API_PUBLIC_KEY:$SERVERLESS_API_PRIVATE_KEY" \
+  --silent \
+  --show-error \
```
kevinAlbs, line 35: Is there a risk that `--show-error` could leak the private key?

kevinAlbs (commented):
This looks fantastic. I only have minor comments and questions.

.evergreen/config.yml
patrickfreed: This is the SCRAM username / password for any cluster created under the drivers instances project (e.g. your `kevin` account or my `patrick` one). In Atlas there are per-project users that automatically get added to each cluster instead of per-cluster users like in a regular deployment. 

I was assuming that all drivers would share the same credentials similar to what we do for the Atlas connectivity tests, but we could easily add accounts per driver if we wished, it would just require more upkeep.

.evergreen/serverless/create-instance.sh
patrickfreed: It's certainly possible, though per [EVG-9754](https://jira.mongodb.org/browse/EVG-9754), all evergreen logs require authentication to be viewed, so the leakage would be limited to inside the company. That being said, we'd have to rotate the private key if it were leaked like that anyways, so I'm still not sure whether it's worth the risk or not to include error diagnostics. Happy to defer to your judgment on this.

.evergreen/config.yml
patrickfreed: Updated. I originally used 16.04 since there was no definition in the config for 18.04, but I just went ahead and added it.

.evergreen/config.yml
patrickfreed: Since we're now using 18.04 I updated this to `os-requires-40`.

.evergreen/config.yml
kevinAlbs: OK, thank you for the explanation. I think we can share an account across all drivers projects to keep it simpler. It does make rotating credentials more frequent (e.g. when someone leaves the company). But that's an existing problem (e.g. we do the same with AWS credentials). There is probably an all-encompassing solution for how we are sharing secrets in evergreen projects.

.evergreen/serverless/create-instance.sh
kevinAlbs: Ah true, this would only be in private logs. Then I'm not too worried, and I'm fine with keeping it.

kevinAlbs (approved):
LGTM!

PR #144, opened 2021-04-13
DRIVERS-555 Add a single-node replset config file

divjotarora:
For DRIVERS-555, we will allow drivers to run the new tests in a separate task against a single-node replica set rather than running them in existing tasks. This is because the tests are timing based and we'd like to reduce the number of variables that could potentially introduce flaky failures (e.g. cumulative load on the server, number of nodes in the cluster, etc). This PR introduces a new config file that can be used by individual drivers to create this cluster.


alcaeus (approved):
LGTM

ShaneHarvey:
Should we also add an ssl+auth variant? Timeouts on TLS connections may work differently depending on the language.

divjotarora:
@ShaneHarvey We're unable to test TLS timeouts in spec tests because there's no infrastructure set up to block TLS handshakes for a specified amount of time. The authentication piece is important because some tests set fail points on `saslContinue` to test timeouts during the MongoDB handshake, but I thought adding TLS would increase the risk of flaky failures due to the additional set of round trips required without providing any benefit.

ShaneHarvey:
Ah tls handshake timeouts are interesting but I was moreso thinking about general timeouts when reading from a TLS connection. I know in python the behavior can differ slightly depending on the tls library.

divjotarora:
@ShaneHarvey I added another config file for single-node replica set with both auth and TLS. I've also updated https://github.com/mongodb/specifications/pull/960 to require testing both with and without TLS enabled in commit https://github.com/mongodb/specifications/pull/960/commits/d6681201a316bebcf4e8fcbdd10bb9949e9592bb.

ShaneHarvey (approved):

PR #145, opened 2021-04-20
NODE-3212: Adding Evergreen Load Balancer Config

durran:
Generates a HAProxy configuration and then runs with two frontends that connect to a single mongos or 2 mongoses.

Usage:

```bash
./load-balancer.sh start
./load-balancer.sh stop
```

Will set `SINGLE_MONGOS_LB_URI` and `MULTI_MONGOS_LB_URI` env variables as the host:port to those load balancers.

TODO:
- [x] Figure out the proper setting of the `MONGODB_URI` for mongo orchestration.

Reference PR of how to run on Evergreen: https://github.com/mongodb/node-mongodb-native/pull/2787

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -329,6 +329,21 @@ functions:
           ${PREPARE_SHELL}
           sh ${DRIVERS_TOOLS}/.evergreen/stop-orchestration.sh
 
+  "start load balancer":
+    - command: shell.exec
+      params:
+        script: |
+          DRIVERS_TOOLS=${DRIVERS_TOOLS} MONGODB_URI=${MONGODB_URI} bash ${DRIVERS_TOOLS}/.evergreen/run-load-balancer.sh start
+    - command: expansions.update
+      params:
+        file: lb-expansion.yml
+
+  "stop load balancer":
+    - command: shell.exec
+      params:
+        script: |
+          DRIVERS_TOOLS=${DRIVERS_TOOLS} bash ${DRIVERS_TOOLS}/.evergreen/run-load-balancer.sh stop
+
   "run tests":
     - command: shell.exec
       type: test
--- a/.evergreen/run-load-balancer.sh
+++ b/.evergreen/run-load-balancer.sh
@@ -1,0 +1,83 @@
+#!/bin/bash
+
+set -o errexit  # Exit the script with error if any of the commands fail
+
+start() {
+  echo "Starting HAProxy..."
+
+  cat <<EOF_HAPROXY_CONFIG >> $DRIVERS_TOOLS/haproxy.conf
+  defaults
+      mode tcp
+      timeout connect 5000ms
+      timeout client 5000ms
+      timeout server 5000ms
+
+  frontend mongos_frontend
+      bind *:8000
+      use_backend mongos_backend
+
+  frontend mongoses_frontend
+      bind *:8001
+      use_backend mongoses_backend
+
+  backend mongos_backend
+      mode tcp
+      server mongos 127.0.0.1:27017 check
+
+  backend mongoses_backend
+      mode tcp
+      server mongos_one 127.0.0.1:27017 check
+      server mongos_two 127.0.0.1:27018 check
+EOF_HAPROXY_CONFIG
+
+  PREFIX=$(echo $MONGODB_URI | grep -Eo "(.*?)@" | cat)
+  SUFFIX=$(echo $MONGODB_URI | grep -Eo "\?(.*)" | cat)
+
+  if [[ $PREFIX = "" ]]
+  then
+    # No auth then just set the URI
+    SINGLE_MONGOS_LB_URI="mongodb://127.0.0.1:8000/"
+    MULTI_MONGOS_LB_URI="mongodb://127.0.0.1:8001/"
+  else
+    # We have auth so append the lb host:port
+    SINGLE_MONGOS_LB_URI="${PREFIX}127.0.0.1:8000/"
+    MULTI_MONGOS_LB_URI="${PREFIX}127.0.0.1:8001/"
+  fi
+
+  if [[ $SUFFIX = "" ]]
+  then
+    # If there are no query params then add only the load balanced option.
+    SINGLE_MONGOS_LB_URI="${SINGLE_MONGOS_LB_URI}?loadBalanced=true"
+    MULTI_MONGOS_LB_URI="${MULTI_MONGOS_LB_URI}?loadBalanced=true"
+  else
+    # If there are query params then append the load balanced option to them.
+    SINGLE_MONGOS_LB_URI="${SINGLE_MONGOS_LB_URI}${SUFFIX}&loadBalanced=true"
+    MULTI_MONGOS_LB_URI="${MULTI_MONGOS_LB_URI}${SUFFIX}&loadBalanced=true"
+  fi
+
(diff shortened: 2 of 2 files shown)


.evergreen/load-balancer.sh
```diff
@@ -0,0 +1,64 @@
...
+      mode tcp
+      server mongos_one 127.0.0.1:27017 check
+      server mongos_two 127.0.0.1:27018 check
+EOF_HAPROXY_CONFIG
+
+  cat $DRIVERS_TOOLS/haproxy.conf
+
+  /usr/sbin/haproxy -D -f $DRIVERS_TOOLS/haproxy.conf -p ./haproxy.pid
+
+  ADDRESS=$(curl -s http://169.254.169.254/latest/meta-data/public-hostname)
```
jyemin, line 37: What's this do?  I tried to run it  on my laptop and it hung.  Why can't we just use `localhost` as the address?

.evergreen/load-balancer.sh
```diff
@@ -0,0 +1,64 @@
...
+      mode tcp
+      server mongos 127.0.0.1:27017 check
+
+  backend mongoses_backend
+      mode tcp
+      server mongos_one 127.0.0.1:27017 check
+      server mongos_two 127.0.0.1:27018 check
+EOF_HAPROXY_CONFIG
+
+  cat $DRIVERS_TOOLS/haproxy.conf
```
jyemin, line 33: Nit: is this just for debugging?  If so, can remove.

.evergreen/load-balancer.sh
```diff
@@ -0,0 +1,64 @@
...
+
+  /usr/sbin/haproxy -D -f $DRIVERS_TOOLS/haproxy.conf -p ./haproxy.pid
+
+  ADDRESS=$(curl -s http://169.254.169.254/latest/meta-data/public-hostname)
+  SINGLE_MONGOS_LB_URI="$ADDRESS:8000"
```
divjotarora, line 38: Related to our offline conversation yesterday: if we expect the underlying sharded cluster to be started with auth/TLS, we either need to put credentials in this URI or otherwise instruct drivers to do so in their test runners. The former is simpler but definitely couples the LB to the cluster because we'd be making assumptions about how the cluster is set up. I'm not sure if there's prior art to go off of here. CC @jyemin maybe you have some thoughts about this?

jyemin: I agree that this URI should contain credentials, as well as `?loadBalanced=true`.

Also, are we planning to run these tests with TLS enabled?  If so we'll have to account for that as well.

alcaeus: Yes, please add the required arguments to the connection string. This avoids ugly workarounds like we're seeing for mongo-orchestration (which doesn't yet expose tls options when required). Ideally, `mongo $SINGLE_MONGOS_LB_URI` should be able to connect and behave as expected.

jyemin: I think it's hard to see how to use this without integration into the Evergreen configuration file.  Can you add a variant that uses this file to start a sharded cluster (with and without auth), a proxy, and runs a test?

.evergreen/load-balancer.sh
durran: This gets the public facing IP for the box on AWS. I thought that would be more useful to export than `localhost`.

divjotarora: I'd prefer `localhost` here so that the script can be run locally as well (currently, it hangs on my laptop too). I'm not sure what benefit we get by using the public IP as DNS resolution is already tested via other test suites. It doesn't seem to add much value to this specific suite.

.evergreen/load-balancer.sh
divjotarora: How would this work if we're testing both with and without auth? Will we have to output two sets of env vars (e.g. `SINGLE_MONGOS_LB_URI` and `SINGLE_MONGOS_LB_AUTH_URI` as well as a similar set for `MULTI_*`) and have the driver pick which ones get plumbed down to the test runner in the Evergreen YAML config?

.evergreen/load-balancer.sh
durran: I've changed this to localhost now.

.evergreen/load-balancer.sh
durran: Removed.

.evergreen/load-balancer.sh
durran: I've updated to include auth and potential parameters, and the Evergreen config reference is here: https://github.com/mongodb/node-mongodb-native/pull/2787

.evergreen/load-balancer.sh
```diff
@@ -0,0 +1,83 @@
...
+  else
+    # If there are query params then append the load balanced option to them.
+    SINGLE_MONGOS_LB_URI="${SINGLE_MONGOS_LB_URI}${SUFFIX}&loadBalanced=true"
+    MULTI_MONGOS_LB_URI="${MULTI_MONGOS_LB_URI}${SUFFIX}&loadBalanced=true"
+  fi
+
+  echo "Single Mongos LB: $SINGLE_MONGOS_LB_URI"
+  echo "Multiple Mongos LB: $MULTI_MONGOS_LB_URI"
+
+  /usr/sbin/haproxy -D -f $DRIVERS_TOOLS/haproxy.conf -p ./haproxy.pid
```
divjotarora, line 61: Should this be `-p $DRIVERS_TOOLS/haproxy.pid`? The `stop` function assumes that's where the pid file is.

durran: Yeah good catch.

.evergreen/load-balancer.sh
```diff
@@ -0,0 +1,86 @@
...
+
+  echo "Single Mongos LB: $SINGLE_MONGOS_LB_URI"
+  echo "Multiple Mongos LB: $MULTI_MONGOS_LB_URI"
+
+  /usr/sbin/haproxy -D -f $DRIVERS_TOOLS/haproxy.conf -p $DRIVERS_TOOLS/haproxy.pid
+
+  echo 'SINGLE_MONGOS_LB_URI: "'$SINGLE_MONGOS_LB_URI'"' > lb-expansion.yml
+  echo 'MULTI_MONGOS_LB_URI: "'$MULTI_MONGOS_LB_URI'"' >> lb-expansion.yml
+
+  export SINGLE_MONGOS_LB_URI="$SINGLE_MONGOS_LB_URI"
```
divjotarora, line 66: nit: I think we can remove these two exports because the exporting is handled by the lb-expansion.yml file now.

divjotarora (approved):
LGTM with one small nit.

.evergreen/load-balancer.sh
durran: Removed.

divjotarora (commented):
Super minor nit: can we rename the file to `run-load-balancer.sh` to match `run-orchestration.sh`?

alcaeus (approved):

durran:
> Super minor nit: can we rename the file to `run-load-balancer.sh` to match `run-orchestration.sh`?

Renamed.

jyemin:
It's a little weird to write: `run-load-balancer.sh stop`.  But I can live with it.

jyemin (approved):
LGTM

PR #146, opened 2021-04-23
Bump idle and connect timeouts in run-load-balancer.sh

divjotarora:
This PR bumps the connect timeout in `run-load-balancer.sh` to 10 seconds, which matches the driver's default for `connectTimeoutMS` and the idle timeouts to 30 minutes, which should be sufficient to ensure that test runners do not encounter unexpected network errors due to severed connections.


jyemin (commented):
LGTM once you've tested it.

jyemin (approved):

PR #148, opened 2021-05-07
GODRIVER-1891 add pkcs8 certs so the go driver can test with the DET …

iwysiu:
…certs

Encrypted generated with password "password" and the command:
`openssl pkcs8 -v2 des3 -topk8 -inform PEM -outform PEM -in client.pem -out client-pkcs8-encrypted.pem`

Unencrypted generated with:
`openssl pkcs8 -topk8 -nocrypt -inform PEM -outform PEM -in client.pem -out client-pkcs8-unencrypted.pem`


benjirewis (approved):
LGTM! How come the `client.pem` from drivers-evergreen-tools is different than our current local copy in the Go driver?

kevinAlbs (approved):
LGTM!

PR #151, opened 2021-05-24
DRIVERS-1603 Drop API version requirement from serverless testing

patrickfreed:
Tested locally and the tests pass without any API version specified, so this PR updates the serverless expansion to no longer include it.


kevinAlbs (approved):
LGTM!

PR #152, opened 2021-05-26
DRIVERS-1603 Update serverless scripts to run on Windows

MikalaiMazurenka:
There are several problems on Windows:
1. `python3` call does not work
2. JSON parser adds a `CRLF` into result for some reason
3. Bash equality comparison `==` does not work

Example fail on Windows: https://evergreen.mongodb.com/task_log_raw/dot_net_driver_serverless_tests__os~windows_64_serverless_tests_patch_cf62d156f72a517c4bb61046ddb23104f58ece0c_60ac0c881e2d173e5cf09858_21_05_24_20_29_13/0?type=T#L181
Patch build with the changes: https://evergreen.mongodb.com/task/dot_net_driver_serverless_tests__os~windows_64_serverless_tests_patch_cf62d156f72a517c4bb61046ddb23104f58ece0c_60adb234e3c3312c2ce547fa_21_05_26_02_29_14

### DIFF
--- a/.evergreen/serverless/create-instance.sh
+++ b/.evergreen/serverless/create-instance.sh
@@ -51,18 +51,24 @@ curl \
 
 echo ""
 
+if [ "Windows_NT" = "$OS" ]; then
+  PYTHON_BINARY=C:/python/Python38/python.exe
+else
+  PYTHON_BINARY=python3
+fi
+
 SECONDS=0
 while [ true ]; do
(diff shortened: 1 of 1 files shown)


patrickfreed (approved):
Just tested it out and it works for me on Linux. Thanks for fixing this!

BTW, I updated the PR title to have DRIVES-1603 instead of the CSHARP ticket so it can be tracked with the rest of the changes to the scripts.

PR #158, opened 2021-07-01
DRIVERS-1603 Switch instance creation to GCP

patrickfreed:
Since the AWS are moving behind a load balancer, we need to temporarily switch to creating the instances on GCP.


kevinAlbs (approved):

alcaeus (approved):
Can confirm that this solves yesterday's instance creation issue for me. Thanks!

PR #161, opened 2021-07-07
DRIVERS-1603 Use private API endpoint in get-instance.sh

patrickfreed:
This PR updates `get-instance.sh` to use the private API endpoint which exposes the full connection string as opposed to just the SRV one as the public endpoint does.


jyemin (approved):
LGTM

PR #163, opened 2021-07-14
Remove mock_kms script in favor of kms_http_server

benjirewis:
Removes the `mock_kms` JS wrapper for `kms_http_server`. It was more complicated to set up and was a trivial wrapper for the underlying `kms_http_server.py`.


.evergreen/csfle/mock_kms.js
```diff
@@ -1,5 +1,18 @@
...
+ * use the following bash snippet.
+ *
+ *   cat <<EOF > kms_setup.json
+ *   {
+ *      "kms_ca_file": "path/to/ca_file",
+ *      "kms_cert_file": "path/to/cert_file",
+ *      "port": "8000"
+ *   }
+ *   EOF
+ *   mongo --nodb mock_kms.js
```
ShaneHarvey, line 14: What value does the `mock_kms.js` script provide? It seems like it's just a trivial wrapper for `kms_http_server.py`.

benjirewis: Yeah... it's actually more complicated to use `mock_kms.js`; I just copied it from the server's code. Removed it in favor of `kms_http_server.py` and added a usage example.

ShaneHarvey (commented):
Two more comments:

1. The .evergreen/csfle/lib/kms_http_server.py → .evergreen/csfle/kms_http_server.py rename is going to break pymongo's tests. I suppose I'm okay because removing mock_kms.js with break Go's tests but it's still unfortunate. pymongo starts the kms servers like this:
   ```
    pushd ${DRIVERS_TOOLS}/.evergreen/csfle
    python -u lib/kms_http_server.py --ca_file ../x509gen/ca.pem --cert_file ../x509gen/expired.pem --port 8000 &
    python -u lib/kms_http_server.py --ca_file ../x509gen/ca.pem --cert_file ../x509gen/wrong-host.pem --port 8001 &
    trap 'kill $(jobs -p)' EXIT HUP
    popd
   ```
 2. How will drivers ensure they run the script in the right environment? From my testing the script needs python >=3.5 with boto3 installed. I think we need to add a venv wrapper script like we did in https://github.com/mongodb-labs/drivers-evergreen-tools/pull/138

benjirewis:
Removing `mock_kms.js` will definitely break Go's tests, but shouldn't be too hard to update. Python will have to drop the `lib/` in front of the `kms_http_server.py` calls, but it seems odd to keep the scripts under `lib` when there's no utility calling them from a parent directory.

I've added a `csfle/activate_venv.sh` script that is almost identical to the one you referenced. I also updated the KMS TLS prose test clarification [PR](https://github.com/mongodb/specifications/pull/1039) to reference the new script.

.evergreen/csfle/activate_venv.sh
```diff
@@ -0,0 +1,25 @@
+set +x
```
ShaneHarvey, line 1: I don't think we want to change the settings in the parent shell do we?

.evergreen/csfle/activate_venv.sh
```diff
@@ -0,0 +1,25 @@
...
+
+if [ "Windows_NT" = "$OS" ]; then
+  PYTHON_BINARY=C:/python/Python38/python.exe
+else
+  PYTHON_BINARY=python3
+fi
```
ShaneHarvey, line 7: Can you change this to:
```
if [ "Windows_NT" = "$OS" ]; then
  PYTHON_BINARY=C:/python/Python38/python.exe
elif [ command -v /opt/python/3.6/bin/python3 ]; then
  PYTHON_BINARY=/opt/python/3.6/bin/python3
else
  PYTHON_BINARY=python3
fi
```

I needed to do this to make this script work on RHEL 7.0:
```
$ python3
-bash: python3: command not found
$ ls /opt/python
2.4  2.5  2.6  2.7  3.2  3.3  3.4  3.5  3.6  jython2.5  jython2.7  pypy  pypy3.5  pypy3.6  toolchain_version
```

ShaneHarvey (changes requested):

.evergreen/csfle/activate_venv.sh
benjirewis: Removed.

.evergreen/csfle/activate_venv.sh
benjirewis: Makes sense, RHEL always has different paths. Thanks for verifying on that distro; changed it to your version.

ShaneHarvey (approved):

PR #167, opened 2021-08-13
Update to 5.0.2, 4.0.26, and 4.4.8

kevinAlbs:
In particular, the C driver requires [SERVER-58794](https://jira.mongodb.org/browse/SERVER-58794), which was fixed in 5.0.2, for passing tests in https://github.com/mongodb/mongo-c-driver/pull/848. The C driver has its own copy. It is not blocked on this PR.

### DIFF
--- a/.evergreen/download-mongodb.sh
+++ b/.evergreen/download-mongodb.sh
@@ -51,7 +51,7 @@ get_mongodb_download_url_for ()
    _DISTRO=$1
    _VERSION=$2
 
-   VERSION_50="5.0.0"
+   VERSION_50="5.0.2"
    VERSION_44="4.4.6"
    VERSION_42="4.2.15"
    VERSION_40="4.0.25"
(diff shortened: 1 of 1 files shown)


.evergreen/download-mongodb.sh
```diff
@@ -51,7 +51,7 @@ get_mongodb_download_url_for ()
    _DISTRO=$1
    _VERSION=$2
 
-   VERSION_50="5.0.0"
+   VERSION_50="5.0.2"
    VERSION_44="4.4.6"
    VERSION_42="4.2.15"
    VERSION_40="4.0.25"
```
ShaneHarvey, line 57: Do you mind updating to 4.0.26 and 4.4.8 as well?

ShaneHarvey (changes requested):

.evergreen/download-mongodb.sh
kevinAlbs: Done

ShaneHarvey (approved):

PR #168, opened 2021-08-23
DRIVERS-1882 update serverless scripts to use AWS

kevinAlbs:
This updates scripts in anticipation of GCP being fronted by a load balancer.
Serverless will no longer offer a test instance without a load balancer.

This adds a LOADBALANCED environment variable to opt-in to switching to a load balanced cluster backed by `AWS`. The rationale for making this opt-in is to prevent causing test failures on drivers that already have serverless tests implemented until they implement changes in DRIVERS-1603.

I will merge this before https://github.com/mongodb/specifications/pull/1053.


.evergreen/serverless/create-instance.sh
```diff
@@ -9,6 +9,17 @@ if [ -z "$PROJECT" ]; then
 fi
 INSTANCE_NAME="$RANDOM-$PROJECT"
 
+# Use the LOADBALANCED environment variable to opt-in to testing
```
kevinAlbs, line 12: If this approach looks good, I can create a DRIVERS ticket as follow up work after DRIVERS-1603 is implemented in language teams to remove the `LOADBALANCED` opt-in environment variable, and always use `AWS` as the backing provider.

patrickfreed (approved):
LGTM!

PR #169, opened 2021-09-09
DRIVERS-1882: Fixed create-instance.sh script so that LOADBALANCED environment variable is ignored when not set.

JamesKovacs:


kevinAlbs (approved):
Thank you for the fix, LGTM!

PR #176, opened 2021-10-27
Explain how srvAddress and mongoURI API response fields are used

jmikola:


patrickfreed (approved):
Thanks for adding these comments. Unfortunately, they will probably go out of date soon given DRIVERS-1967, though I think they're still worth having in until then.

PR #180, opened 2021-11-29
Address miscellaneous Python-related CSE Evergreen test failures

eramongodb:
During development of CDRIVER-4225, a couple CSE task failures on Evergreen due to Python environment setup failure were encountered. This PR introduces changes that resolve some of these failures.

1. Add server toolchain python3 to list of candidates

The zSeries variant does not have a python3 executable available in any of the existing set of paths inspected. [This commit](https://github.com/mongodb-labs/drivers-evergreen-tools/commit/c5fd810d5f5f7b95c49a1fdda670f244062907ad) adds the python3 executable provided by the server toolchain (already being used for GCC executable) as a last resort to the list of candidates.

2. Pin cryptography to permit use of `CRYPTOGRAPHY_DONT_BUILD_RUST`

The `cryptography` package was [updated to require the Rust toolchain as a dependency](https://github.com/pyca/cryptography/issues/5771), breaking support for numerous platforms and configurations, including ours for the zSeries and Power8 Evergreen variants. [This workaround](https://github.com/mongodb-labs/drivers-evergreen-tools/commit/1941d9225197867242165651094eaa6e253337c7) pins the `cryptography` package version to the latest version that still supported the `CRYPTOGRAPHY_DONT_BUILD_RUST` environment variable, which resolves issues on the zSeries variant. The Power8 variant continues to fail due to `ModuleNotFoundError: No module named '_cffi_backend'`.

### DIFF
--- a/.evergreen/csfle/activate_venv.sh
+++ b/.evergreen/csfle/activate_venv.sh
@@ -13,7 +13,7 @@ fi
 # create venv on first run
 if [ ! -d kmstlsvenv ]; then
    FIRST_RUN=1
-   ${PYTHON_BINARY} -m virtualenv -p ${PYTHON_BINARY} kmstlsvenv
+   virtualenv -p ${PYTHON_BINARY} kmstlsvenv
 fi
 
 # always activate venv
--- a/.evergreen/csfle/activate_venv.sh
+++ b/.evergreen/csfle/activate_venv.sh
@@ -25,5 +25,5 @@ fi
 
 # install dependencies on first run
 if [ ! -z $FIRST_RUN ]; then
-  pip install --upgrade boto3~=1.19 pykmip~=0.10.0
+  CRYPTOGRAPHY_DONT_BUILD_RUST=1 pip install --upgrade boto3~=1.19 cryptography~=3.4.8 pykmip~=0.10.0
 fi
--- a/.evergreen/csfle/activate_venv.sh
+++ b/.evergreen/csfle/activate_venv.sh
@@ -2,14 +2,18 @@ if [ "Windows_NT" = "$OS" ]; then
   PYTHON_BINARY=C:/python/Python38/python.exe
 elif command -v /opt/python/3.6/bin/python3; then
   PYTHON_BINARY=/opt/python/3.6/bin/python3
-else
+elif command -v python3; then
   PYTHON_BINARY=python3
+elif command -v /opt/mongodbtoolchain/v2/bin/python3; then
+  PYTHON_BINARY=/opt/mongodbtoolchain/v2/bin/python3
+else
+  echo "error: unable to find a supported python3 executable"
 fi
 
 # create venv on first run
 if [ ! -d kmstlsvenv ]; then
    FIRST_RUN=1
-   virtualenv -p ${PYTHON_BINARY} kmstlsvenv
+   ${PYTHON_BINARY} -m virtualenv -p ${PYTHON_BINARY} kmstlsvenv
 fi
 
 # always activate venv


.evergreen/csfle/activate_venv.sh
```diff
@@ -2,14 +2,18 @@ if [ "Windows_NT" = "$OS" ]; then
...
+  PYTHON_BINARY=/opt/mongodbtoolchain/v2/bin/python3
+else
+  echo "error: unable to find a supported python3 executable"
 fi
 
 # create venv on first run
 if [ ! -d kmstlsvenv ]; then
    FIRST_RUN=1
-   virtualenv -p ${PYTHON_BINARY} kmstlsvenv
+   ${PYTHON_BINARY} -m virtualenv -p ${PYTHON_BINARY} kmstlsvenv
```
ShaneHarvey, line 16: Is this change needed? I'm concerned it may break some platforms.

eramongodb: How could this break existing behavior given the remainder of the script assumes/requires a valid `${PYTHON_BINARY}`? Are there platforms that install `virtualenv` without a `python3` binary?

.evergreen/csfle/activate_venv.sh
```diff
@@ -21,5 +25,5 @@ fi
 
 # install dependencies on first run
 if [ ! -z $FIRST_RUN ]; then
-  pip install --upgrade boto3~=1.19 pykmip~=0.10.0
-fi
\ No newline at end of file
+  CRYPTOGRAPHY_DONT_BUILD_RUST=1 pip install --upgrade boto3~=1.19 cryptography~=3.4.8 pykmip~=0.10.0
```
benjirewis, line 28: Out of curiosity, where is the new `CRYPTOGRAPHY_DONT_BUILD_RUST` env variable referenced?

.evergreen/csfle/activate_venv.sh
benjirewis: I also am not sure how this would break existing behavior, but I do wonder if it's redundant with the following `-p {PYTHON_BINARY}`.

.evergreen/csfle/activate_venv.sh
eramongodb: It is used during build of the `cryptography` package; see [cryptography 3.4 documentation](https://cryptography.io/en/3.4/faq.html#installing-cryptography-fails-with-error-can-not-find-rust-compiler) and the [latest docs](https://github.com/pyca/cryptography/blob/75be92de8e3bce9adcec42ef3967bed0d4500902/docs/installation.rst#rust) regarding Rust requirement. I don't expect this workaround (version pin + `CRYPTOGRAPHY_DONT_BUILD_RUST`) to be a long-term solution; I think the proper long-term solution would be to install the Rust toolchain on all necessary Evergreen variants, but that is beyond the scope of changes I require to unblock related work.

.evergreen/csfle/activate_venv.sh
ShaneHarvey: In the past there were platforms where `virtualenv` exists but `python3 -m virtualenv` does not. I'm not 100% sure the problem still exists but if this change isn't required I think it's safer to remove it.

eramongodb: @benjirewis It's not quite the same: `${PYTHON_BINARY} -m virtualenv` requests a `virtualenv` binary installed/associated with `${PYTHON_BINARY}`, whereas just `virtualenv` uses the system-provided binary, which may not be the same as that associated with `${PYTHON_BINARY}` (e.g. a Python2 binary, or a standalone binary installed via packages?). I suppose I was assuming too much to think that the Python binary and associated packages would be consistent on all platforms.

@ShaneHarvey If that is the case, I will undo this change, as I believe it wasn't strictly necessary to unblock related work.

.evergreen/csfle/activate_venv.sh
benjirewis: SGTM; thanks for the context.

benjirewis (approved):
LGTM

ShaneHarvey (approved):

kevinAlbs (approved):

kevinAlbs (approved):

PR #181, opened 2021-12-03
DRIVERS-1357: Add socks5srv.py helper script

addaleax:
- Spec PR: https://github.com/mongodb/specifications/pull/1103
- Node.js driver PR that makes use of this script:
  https://github.com/mongodb/node-mongodb-native/pull/3041

### DIFF
--- a/.evergreen/socks5srv.py
+++ b/.evergreen/socks5srv.py
@@ -6,7 +6,6 @@ import re
 import argparse
 
 # Usage: python3 socks5srv.py --port port [--auth username:password] [--map 'host:port to host:port' ...]
-# TODO: Move this script into the shared drivers-evergreen-tools repository
 
 class AddressRemapper:
   """A helper for remapping (host, port) tuples to new (host, port) tuples
@@ -103,13 +102,20 @@ class Socks5Handler(socketserver.BaseRequestHandler):
         return None
       assert len(n) == 1
       n = n[0]
-    result = b''
-    while len(result) < n:
-      buf = self.request.recv(n - len(result))
-      if buf == b'':
-        return None
-      result += buf
-    return result
+
+    buf = bytearray(n)
+    mv = memoryview(buf)
+    bytes_read = 0
+    while bytes_read < n:
+        try:
+            chunk_length = self.request.recv_into(mv[bytes_read:])
+        except OSError as exc:
+            return None
+        if chunk_length == 0:
+            return None
+
+        bytes_read += chunk_length
+    return bytes(buf)
 
   def create_outgoing_tcp_connection(self, dst, port):
     """Create an outgoing TCP connection to dst:port"""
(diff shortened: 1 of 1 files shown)


nbbeeken (approved):

durran (approved):

.evergreen/socks5srv.py
```diff
@@ -0,0 +1,237 @@
+#!/usr/bin/env python3
+import socketserver
+import socket
+import select
+import re
+import argparse
+
+# Usage: python3 socks5srv.py --port port [--auth username:password] [--map 'host:port to host:port' ...]
+# TODO: Move this script into the shared drivers-evergreen-tools repository
```
ShaneHarvey, line 9: Remove this todo.

.evergreen/socks5srv.py
```diff
@@ -0,0 +1,237 @@
...
+      if len(n) == 0:
+        return None
+      assert len(n) == 1
+      n = n[0]
+    result = b''
+    while len(result) < n:
+      buf = self.request.recv(n - len(result))
+      if buf == b'':
+        return None
+      result += buf
```
ShaneHarvey, line 111: So building the result buffer like this `+=` can yield o(n^2) performance. See PYTHON-1513. I'm not sure if this method is only used for reading small pieces of data but a more efficient way would be:
```python
    buf = bytearray(n)
    mv = memoryview(buf)
    bytes_read = 0
    while bytes_read < n:
        try:
            chunk_length = self.request.recv_into(mv[bytes_read:])
        except OSError as exc:
            return None
        if chunk_length == 0:
            return None

        bytes_read += chunk_length
    return bytes(buf)
```

.evergreen/socks5srv.py
```diff
@@ -0,0 +1,237 @@
...
+        outgoing.close()
+        continue
+      break
+    return outgoing
+
+  def handle(self):
+    """Handle the Socks5 communication with a freshly connected client"""
+
+    # Client greeting
+    if self.request.recv(1) != b'\x05': # Socks5 only
```
ShaneHarvey, line 136: Any request.recv/sendall call can raise an OSError exception, do you want to handle those explicitly or just let the thread fail with an exception? I think the latter is fine.

.evergreen/socks5srv.py
```diff
@@ -0,0 +1,237 @@
...
+      return
+    n_auth = self.request.recv(1)
+    client_auth_methods = self.read_exact(n_auth)
+    if client_auth_methods is None:
+      return
+
+    # choose either no-auth or username/password
```
ShaneHarvey, line 143: Some links to the sock5 spec you used to write this handshake protocol might be helpful for future maintainers.

.evergreen/socks5srv.py
```diff
@@ -0,0 +1,237 @@
...
+
+    self.raw_proxy(self.request, outgoing)
+
+  def raw_proxy(self, a, b):
+    """Proxy data between sockets a and b as-is"""
+
+    with a, b:
+      while True:
+        try:
+          (readable, _, _) = select.select([a, b], [], [])
```
ShaneHarvey, line 213: select will fail if the fd is > 1024 so we should prefer [poll](https://docs.python.org/3/library/select.html#poll-objects) but since this is test only it's not a big deal. We're not planning to test this with >1000 connections are we?

ShaneHarvey (changes requested):

.evergreen/socks5srv.py
addaleax: Applied your suggestion, but yeah, I wouldn’t have been too worried about that here – the buffer size is limited by the procotol definition to 255 bytes, and the actual usage in the tests doesn’t even come close to that..

.evergreen/socks5srv.py
addaleax: Yeah, I’d also say so – I don’t really see a better alternative and I think this behavior is fine for our usage.

.evergreen/socks5srv.py
addaleax: Also here, I agree that this is just fine for our usage – we are not planning to run this with more than 1000 connections at a time for sure :)

ShaneHarvey (approved):
LGTM!

PR #183, opened 2022-01-04
feat(DRIVERS-1983): add load balancer port in config

durran:
Adds new mongo orchestration configurations for sharded clusters when using a load balancer to specify the load balancer port config option to the mongoses. Requires that a new `LOAD_BALANCER` environment variable is set (to any value) to load the new configs and will configure the mongoses to ports 27050 and 27051, and essentially replaces the old `FAKE_MONGODB_SERVICE_ID` that drivers can now remove and run the lb tests against "latest". Updates the haproxy configuration when running the load balancer to point at these new ports.

Node implementation: https://github.com/mongodb/node-mongodb-native/pull/3089

Current Node patch build: https://spruce.mongodb.com/version/61f09db81e2d1776cfb6c7cf/tasks

### DIFF
--- a/.evergreen/start-orchestration.sh
+++ b/.evergreen/start-orchestration.sh
@@ -40,7 +40,7 @@ elif $PYTHON -m virtualenv --system-site-packages --never-download venv || virtu
 fi
 
 # Install from github to get the latest mongo-orchestration.
-pip install --upgrade 'https://github.com/10gen/mongo-orchestration/archive/DRIVERS-1983.tar.gz'
+pip install --upgrade 'https://github.com/mongodb/mongo-orchestration/archive/master.tar.gz'
 pip list
 cd -
 
--- a/.evergreen/start-orchestration.sh
+++ b/.evergreen/start-orchestration.sh
@@ -40,7 +40,7 @@ elif $PYTHON -m virtualenv --system-site-packages --never-download venv || virtu
 fi
 
 # Install from github to get the latest mongo-orchestration.
-pip install --upgrade 'https://github.com/mongodb/mongo-orchestration/archive/master.tar.gz'
+pip install --upgrade 'https://github.com/10gen/mongo-orchestration/archive/DRIVERS-1983.tar.gz'
 pip list
 cd -
 
--- a/.evergreen/orchestration/configs/sharded_clusters/load-balancer-ssl.json
+++ b/.evergreen/orchestration/configs/sharded_clusters/load-balancer-ssl.json
@@ -1,0 +1,56 @@
+{
+  "id": "shard_cluster_1",
+  "shards": [
+    {
+      "id": "sh01",
+      "shardParams": {
+        "members": [
+          {
+            "procParams": {
+              "ipv6": true,
+              "bind_ip": "127.0.0.1,::1",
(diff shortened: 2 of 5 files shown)


.evergreen/start-orchestration.sh
```diff
@@ -40,7 +40,7 @@ elif $PYTHON -m virtualenv --system-site-packages --never-download venv || virtu
 fi
 
 # Install from github to get the latest mongo-orchestration.
-pip install --upgrade 'https://github.com/mongodb/mongo-orchestration/archive/master.tar.gz'
+pip install --upgrade 'https://github.com/10gen/mongo-orchestration/archive/DRIVERS-1983.tar.gz'
```
durran, line 43: This will need to be reverted before merging.

ShaneHarvey (approved):
LGTM. Is this change backwards compatible or will drivers need to update their load balancer test set up?

ShaneHarvey:
Also the [Node patch build](https://spruce.mongodb.com/task/mongo_node_driver_next_ubuntu_18.04_erbium_test_load_balancer_patch_54f23529bcc9dab401ae1e48356e2af5fd9673e5_61f0390561837d380d32b022_22_01_25_17_53_10/logs?execution=0) is failing. Is that expected?

durran:
> Also the [Node patch build](https://spruce.mongodb.com/task/mongo_node_driver_next_ubuntu_18.04_erbium_test_load_balancer_patch_54f23529bcc9dab401ae1e48356e2af5fd9673e5_61f0390561837d380d32b022_22_01_25_17_53_10/logs?execution=0) is failing. Is that expected?

Yes I'm working on the changes to the test runner there as the server implementation was not as originally expected and there will probably be runner changes for all drivers, but small.

> LGTM. Is this change backwards compatible or will drivers need to update their load balancer test set up?

Drivers will need to do small changes. Remove the use of `FAKE_MONGODB_SERVICE_ID` in the code, add the `LOAD_BALANCER` env var to their env when running lb tests, and maybe small runner changes, but nothing major.

ShaneHarvey:
Great I guess one last thing. I assume this change means that we can no longer test with 5.0? Are we okay with that change or is there a requirement for drivers to keep testing 5.0?

durran:
> Great I guess one last thing. I assume this change means that we can no longer test with 5.0? Are we okay with that change or is there a requirement for drivers to keep testing 5.0?

For now drivers will need to test against "latest" in the evergreen waterfall, until 5.2.0 is GA. Then any version greater than or equal to 5.2.0 - I didn't see it get backported into 5.0.x.

ShaneHarvey (approved):
LGTM. It would be good to note in Jira+slack that this change will break driver tests on 5.0. It's for sure going to break python's test suite.

durran:
> LGTM. It would be good to note in Jira+slack that this change will break driver tests on 5.0. It's for sure going to break python's test suite.

Should I hold off merging then? Since the update to the latest drivers tools will break everything - maybe give time to prepare?

ShaneHarvey:
I think it will be easier for teams to implement these changes if you merge otherwise they'll need to pull drivers-evergreen-tools from a branch (like you've done in your node PR).

PR #194, opened 2022-04-19
DRIVERS-1967: Serverless testing uses load balancer fronting single proxy

jmikola:
https://jira.mongodb.org/browse/DRIVERS-1967

Updates serverless scripts to use a multi-tenant MongoDB (MTM) deployment with the SINGLE_TARGET_SERVERLESS_DEPLOYMENT feature flag enabled (CLOUDP-117288).

This simplifies much of the original logic since there is now only a single SRV URI. For BC, create-instance.sh still outputs all of the original expansions. Those can be cleaned up at a later date after all downstream drivers migrate to the new expansions.

The create-instance.sh script was also updated to solicit an explicit instance name (e.g. generated within the Evergreen config).

### DIFF
--- a/.evergreen/serverless/create-instance.sh
+++ b/.evergreen/serverless/create-instance.sh
@@ -1,21 +1,20 @@
 #!/bin/bash
 
 set -o errexit
-set +o xtrace # disable xtrace to ensure credentials aren't leaked
+set +o xtrace # Disable xtrace to ensure credentials aren't leaked
 
-# INSTANCE_NAME has the following limitations:
-# https://docs.atlas.mongodb.com/reference/atlas-limits/#label-limits
-INSTANCE_NAME="$RANDOM-DRIVERTEST"
-
-# Set the LOADBALANCED environment variable to "ON" to opt-in to
-# testing load balanced serverless instances.
-if [ "$LOADBALANCED" = "ON" ]; then
-    BACKING_PROVIDER_NAME="AWS"
-    INSTANCE_REGION_NAME="US_EAST_1"
-else
-    BACKING_PROVIDER_NAME="GCP"
-    INSTANCE_REGION_NAME="CENTRAL_US"
-fi
+# Supported environment variables:
+#
+#   SERVERLESS_INSTANCE_NAME    Optional. Serverless instance to create (defaults to a random name).
+#   SERVERLESS_DRIVERS_GROUP    Required. Atlas group for drivers testing.
+#   SERVERLESS_API_PUBLIC_KEY   Required. Public key for Atlas API request.
+#   SERVERLESS_API_PRIVATE_KEY  Required. Private key for Atlas API request.
+#
+# On success, this script will output serverless-expansion.yml with the
+# following expansions:
+#
+#   SERVERLESS_URI            SRV connection string for newly created instance
+#   SERVERLESS_INSTANCE_NAME  Name of newly created instance (required for "get" and "delete" scripts)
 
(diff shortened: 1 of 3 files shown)


.evergreen/serverless/create-instance.sh
```diff
@@ -1,21 +1,20 @@
...
+#
+#   SERVERLESS_INSTANCE_NAME    Optional. Serverless instance to create (defaults to a random name).
+#   SERVERLESS_DRIVERS_GROUP    Required. Atlas group for drivers testing.
+#   SERVERLESS_API_PUBLIC_KEY   Required. Public key for Atlas API request.
+#   SERVERLESS_API_PRIVATE_KEY  Required. Private key for Atlas API request.
+#
+# On success, this script will output serverless-expansion.yml with the
+# following expansions:
+#
+#   SERVERLESS_URI            SRV connection string for newly created instance
```
jmikola, line 16: I considered using `MONGODB_URI`, but this prefix is consistent `SERVERLESS_INSTANCE_NAME` and removes ambiguity. The previous version of this script set many environment variables (e.g. `MONGODB_URI`, `MONGODB_SRV_URI`, `SINGLE_ATLASPROXY_SERVERLESS_URI`), so changing to something new seemed like a clean break (even though we still define the old vars for BC).

patrickfreed (approved):

.evergreen/serverless/create-instance.sh
```diff
@@ -46,94 +61,59 @@ curl \
...
-print('SERVERLESS: serverless')
-print('MULTI_ATLASPROXY_SERVERLESS_URI: "%s"' % (multi_atlasproxy_serverless_uri))
-print('SINGLE_ATLASPROXY_SERVERLESS_URI: "%s"' % (single_atlasproxy_serverless_uri))
+        echo "Setup done! ($(($duration / 60))m $(($duration % 60))s elapsed)"
+
+        SERVERLESS_URI=`echo $API_RESPONSE | $PYTHON_BINARY -c "import sys, json; print(json.load(sys.stdin)['connectionStrings']['standardSrv'])" | tr -d '\r\n'`
+
+        SERVERLESS_URI=$SERVERLESS_URI \
+        SERVERLESS_INSTANCE_NAME=$SERVERLESS_INSTANCE_NAME \
+        echo << EOF | tee serverless-expansion.yml
```
kevinAlbs, line 100:
```suggestion
        cat << EOF | tee serverless-expansion.yml
```

Running `create-instance.sh` in zsh locally outputs an empty serverless-expansion.yml.

kevinAlbs (commented):
Looks great! The additional commentary in the script and "Steps taken to configure a new cloud-dev project" in the ticket comments are a really helpful reference. 

Rerunning the Go driver serverless task with the existing `SERVERLESS_DRIVERS_GROUP` sees an empty `serverless-expansion.yml`. Changing `echo` to `cat` as suggested may resolve that.

.evergreen/serverless/create-instance.sh
jmikola: Can do. If we're using `cat`, though, I expect there's no reason to use `tee`.

Also curious how you ran this in `zsh` if the script starts with `#!/bin/bash`. Were you explicitly running it through `zsh` to overcome that?

jmikola: This should now be fixed. Feel free to test/resolve.

kevinAlbs (approved):

PR #204, opened 2022-06-01
DRIVERS-2346 Use venv instead of virtualenv

ShaneHarvey:
Attempting to fix https://jira.mongodb.org/browse/DRIVERS-2346

Testing here: https://spruce.mongodb.com/version/6297f4961e2d1748f0c12306

### DIFF
--- a/.evergreen/csfle/activate_venv.sh
+++ b/.evergreen/csfle/activate_venv.sh
@@ -18,6 +18,9 @@ fi
 
 # always activate venv
 if [ "Windows_NT" = "$OS" ]; then
+  # Workaround https://bugs.python.org/issue32451:
+  # kmstlsvenv/Scripts/activate: line 3: $'\r': command not found
+  dos2unix kmstlsvenv/Scripts/activate || true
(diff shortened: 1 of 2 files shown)


.evergreen/csfle/activate_venv.sh
```diff
@@ -13,11 +13,14 @@ fi
...
    FIRST_RUN=1
-   virtualenv -p ${PYTHON_BINARY} kmstlsvenv
+   ${PYTHON_BINARY} -m venv kmstlsvenv
 fi
 
 # always activate venv
 if [ "Windows_NT" = "$OS" ]; then
+  # Workaround https://bugs.python.org/issue32451:
+  # kmstlsvenv/Scripts/activate: line 3: $'\r': command not found
+  dos2unix kmstlsvenv/Scripts/activate || true
```
ShaneHarvey, line 23: We use this same approach in pymongo's test suite here: https://github.com/mongodb/mongo-python-driver/blob/75685c006c9184230af131a09611aa3e6e1ac649/.evergreen/utils.sh#L23-L25

PR #205, opened 2022-06-02
DRIVERS-2346 Upgrade pip/setuptools when using venv

ShaneHarvey:
Testing here: https://spruce.mongodb.com/version/6298eb10c9ec447522cc39db/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/.evergreen/utils.sh
+++ b/.evergreen/utils.sh
@@ -5,10 +5,10 @@
 venvcreate () {
     PYTHON="$1"
     VENVPATH="$2"
-    if $PYTHON -m venv -h>/dev/null; then
-        VENV="$PYTHON -m venv"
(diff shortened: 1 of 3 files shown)


.evergreen/utils.sh
```diff
@@ -0,0 +1,36 @@
+# Usage:
+# venvcreate /path/to/python /output/path/for/venv
+# * param1: Python binary to use for the virtualenv
+# * param2: Path to the virtualenv to create
+venvcreate () {
```
ShaneHarvey, line 5: Note I copied this from pymongo (with some minor changes): https://github.com/mongodb/mongo-python-driver/blob/4.1.0/.evergreen/utils.sh#L5-L33

PR #216, opened 2022-07-20
DRIVERS-2377 Add scripts for GCP KMS tests

kevinAlbs:
# Summary
- Add scripts to create a temporary Google Compute Engine VM to test [DRIVERS-2377](https://jira.mongodb.org/browse/DRIVERS-2377).

# Background & Motivation

These scripts are used for integration testing on GCE for [DRIVERS-2377](https://jira.mongodb.org/browse/DRIVERS-2377).

The Go implementation (https://github.com/mongodb/mongo-go-driver/pull/1029) tests these scripts. The [testgcpkms_taskgroup](https://github.com/mongodb/mongo-go-driver/pull/1029/files#diff-2bc841e86ce96b7b422ae203fd8315d0b2a461956cecbe0e096420656fc3fb12R2242) has an example of use.

These scripts are a short-term solution to testing on GCP.

- [BUILD-15384](https://jira.mongodb.org/browse/BUILD-15384) proposes installing `gcloud` on necessary distros. That should enable removing the download-gcloud.sh script.
- [BUILD-15361](https://jira.mongodb.org/browse/BUILD-15361) proposes a longer term solution. This may enable removing these scripts entirely.


.evergreen/csfle/gcpkms/create-and-setup-instance.sh
```diff
@@ -0,0 +1,44 @@
...
+# create-instance.sh sets INSTANCENAME.
+echo "create-instance.sh ... begin"
+. $GCPKMS_DRIVERS_TOOLS/.evergreen/csfle/gcpkms/create-instance.sh
+echo "create-instance.sh ... end"
+
+# Sleep for 60 seconds for VM to finish booting.
+# Otherwise SSH may fail. See https://cloud.google.com/compute/docs/troubleshooting/troubleshooting-ssh.
+echo "Sleeping for 60 seconds ... begin"
+sleep 60
+echo "Sleeping for 60 seconds ... end"
```
vector-of-bool, line 35: Can this be done by checking in a loop? i.e. continually "try to connect until it is available, or abort after N minutes"

Sleeping will introduce unnecessary latency for fast boots, and spuriously fail for slow boots.

.evergreen/csfle/gcpkms/create-and-setup-instance.sh
```diff
@@ -0,0 +1,44 @@
...
+
+echo "setup-instance.sh ... begin"
+. $GCPKMS_DRIVERS_TOOLS/.evergreen/csfle/gcpkms/setup-instance.sh
+echo "setup-instance.sh ... end"
+
+echo "GCPKMS_GCLOUD: $GCPKMS_GCLOUD" > testgcpkms-expansions.yml
+echo "GCPKMS_INSTANCENAME: $GCPKMS_INSTANCENAME" >> testgcpkms-expansions.yml
+echo "GCPKMS_PROJECT: $GCPKMS_PROJECT" >> testgcpkms-expansions.yml
+echo "GCPKMS_ZONE: $GCPKMS_ZONE" >> testgcpkms-expansions.yml
```
vector-of-bool, line 44: I've seen this pattern a few times and it often makes me wary. Instead of building a YAML file by hand, could we use a tiny Python script:

```bash
python -c 'import is, json; print(json.dumps(os.environ))' > env.json
```

Then all environment variables will be available as later expansions without needing to list them and ensure they are properly escaped for YAML.

.evergreen/csfle/gcpkms/create-and-setup-instance.sh
kevinAlbs: Good idea. Done.

.evergreen/csfle/gcpkms/create-and-setup-instance.sh
kevinAlbs: I consider the expansions to be the output of this script. Drivers are expected to load the expansions and export them as environment variables for later scripts. Here is the [example in Go](https://github.com/mongodb/mongo-go-driver/pull/1029/files#diff-2bc841e86ce96b7b422ae203fd8315d0b2a461956cecbe0e096420656fc3fb12R1979-R1982).

IMO adding all environment variables is unnecessary. It may risk that drivers rely on expansions that were not intended to be produced by this script.

vector-of-bool (approved):

.evergreen/csfle/gcpkms/remote-scripts/startup.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+# Delete the GCE instance after a period of time.
+# Refer: https://cloud.google.com/community/tutorials/create-a-self-deleting-virtual-machine
+sleep 7200
```
ShaneHarvey, line 5: Do we always leave the instance alive for 2 hours? Can we delete it sooner when the EVG task completes?

kevinAlbs: I expect drivers to delete as soon as the task finishes with a `teardown_group`. Here is the [example](https://github.com/mongodb/mongo-go-driver/pull/1029/files#diff-2bc841e86ce96b7b422ae203fd8315d0b2a461956cecbe0e096420656fc3fb12R2280) with the Go implementation. The two hour limit is a fallback in case instances are leftover during development, or due to an unforeseen error.

.evergreen/csfle/gcpkms/mock_server.py
```diff
@@ -0,0 +1,90 @@
...
+
+    resp = requests.post(url="https://oauth2.googleapis.com/token", data={
+        "grant_type": "urn:ietf:params:oauth:grant-type:jwt-bearer",
+        "assertion": assertion
+    })
+
+    if resp.status_code != 200:
+        msg = textwrap.dedent("""
+        Unexpected non-200 status.
+        Got status {resp.status_code} on HTTP response:
```
ShaneHarvey, line 61: Is this suppose to be a f-string?

.evergreen/csfle/gcpkms/remote-scripts/startup.sh
ShaneHarvey: SGTM

ShaneHarvey (approved):

.evergreen/csfle/gcpkms/mock_server.py
kevinAlbs: Good catch. Added missing `f`.

PR #221, opened 2022-08-01
CSHARP-4254: Test with consistent versions of crypt_shared and server. 

DmitryLukyanov:
Related driver PR https://github.com/mongodb/mongo-csharp-driver/pull/860

### DIFF
--- a/.evergreen/download-mongodb.sh
+++ b/.evergreen/download-mongodb.sh
@@ -579,7 +579,7 @@ download_and_extract_crypt_shared ()
    cd ..
    rm -rf crypt_shared_download
 
-   RELATIVE_CRYPT_SHARED_LIB_PATH="$(find -type f \( -name "mongo_crypt_v1.dll" -o -name "mongo_crypt_v1.so" -o -name "mongo_crypt_v1.dylib" \))"
+   RELATIVE_CRYPT_SHARED_LIB_PATH="$(find . -type f \( -name "mongo_crypt_v1.dll" -o -name "mongo_crypt_v1.so" -o -name "mongo_crypt_v1.dylib" \))"
    eval $__CRYPT_SHARED_LIB_PATH=$DRIVERS_TOOLS/../$(basename $RELATIVE_CRYPT_SHARED_LIB_PATH)
 
 }
--- a/.evergreen/download-mongodb.sh
+++ b/.evergreen/download-mongodb.sh
@@ -565,7 +565,7 @@ download_and_extract_crypt_shared ()
 {
    MONGO_CRYPT_SHARED_DOWNLOAD_URL=$1
    EXTRACT=$2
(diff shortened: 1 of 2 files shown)


.evergreen/download-mongodb.sh
```diff
@@ -565,16 +565,22 @@ download_and_extract_crypt_shared ()
 {
    MONGO_CRYPT_SHARED_DOWNLOAD_URL=$1
    EXTRACT=$2
+   __CRYPT_SHARED_LIB_PATH=$3
```
kevinAlbs, line 568:
```suggestion
   __CRYPT_SHARED_LIB_PATH=${3:-CRYPT_SHARED_LIB_PATH}
```

Suggested change is to avoid breaking existing tests. The current Go driver scripts [call with two arguments](https://github.com/mongodb/mongo-go-driver/blob/5073f14299701635e23a2c3c5dd3f14c37f38e03/.evergreen/config.yml#L211.). They will be updated. But in case there are other callers.

kevinAlbs (approved):
Thank you for the change. I think this makes it much easier to get the shared library path.

I left one suggested change to avoid breaking tests relying on existing behavior.

.evergreen/run-orchestration.sh
```diff
@@ -102,6 +102,19 @@ echo 'MONGODB_URI: "'$URI'"' > mo-expansion.yml
...
 
+if [ -z $MONGO_CRYPT_SHARED_DOWNLOAD_URL ]; then
+  echo "There is no crypt_shared library for distro='$DISTRO' and version='$MONGODB_VERSION'".
+else
+  echo "Downloading crypt_shared package from $MONGO_CRYPT_SHARED_DOWNLOAD_URL"
+  download_and_extract_crypt_shared "$MONGO_CRYPT_SHARED_DOWNLOAD_URL" "$EXTRACT" CRYPT_SHARED_LIB_PATH
+  echo "CRYPT_SHARED_LIB_PATH:" $CRYPT_SHARED_LIB_PATH
+
+cat <<EOT >> mo-expansion.yml
+CRYPT_SHARED_LIB_PATH: "$CRYPT_SHARED_LIB_PATH"
```
BorisDog, line 113: In .Net EG scripts we had a bug where the shared lib was downloaded, but CRYPT_SHARED_LIB_PATH was not passed through. So we added an explicit validation for that (which is being removed with this change).

Should we validate that CRYPT_SHARED_LIB_PATH is set here (when shared lib is requested)? Or maybe validate that in our .Net scripts? Or not worry about it at all?
I am afraid that due to some future bug/misconfiguration, we might silently fallback to mongocryptd, and that might go unnoticed for quite a while.
cc @kevinAlbs

DmitryLukyanov: I'm ok to add validation on this level, will do it tomorrow

PR #225, opened 2022-08-09
DRIVERS-2377 Specify a non-root ssh user

kevinAlbs:
# Summary
- Specify a non-root ssh user for GCP script.
- Capture output of last failed ssh attempt.

# Background & Motivation

The GCE `debian-11` instance disables root login over ssh. The contents of `/etc/ssh/sshd_config` contain `PermitRootLogin no`.
The default user on the Evergreen distro `rhel70-small` is `root`. This appears to be the cause of observed [task failures on the ssh command](https://spruce.mongodb.com/task/mongo_java_driver_testgcpkms_variant_testgcpkms_task_patch_6a2389083a9b744bb771fdf2ce767d84adda7803_62f256f10ae606469f19ddb6_22_08_09_12_45_37/logs?execution=0) with the message `Permission denied (publickey)`.

### DIFF
--- a/.evergreen/csfle/gcpkms/create-and-setup-instance.sh
+++ b/.evergreen/csfle/gcpkms/create-and-setup-instance.sh
@@ -32,14 +32,14 @@ echo "create-instance.sh ... end"
 # Otherwise SSH may fail. See https://cloud.google.com/compute/docs/troubleshooting/troubleshooting-ssh.
 wait_for_server () {
     for i in $(seq 300); do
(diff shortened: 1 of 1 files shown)


jyemin (commented):
Where is the gcpkms user created?

kevinAlbs:
> Where is the gcpkms user created?

IIUC the `gcloud` command creates the SSH public key with the remote user if needed. From `gcloud compute ssh --help`:

> gcloud compute ssh ensures that the user's public SSH key is present in the
    project's metadata. If the user does not have a public SSH key, one is
    generated using ssh-keygen(1) (if the --quiet flag is given, the generated
    key will have an empty passphrase).

jyemin (approved):

PR #226, opened 2022-08-10
Update MongoDB Rapid Version

juliusgeo:


ShaneHarvey:
I just realized that https://github.com/mongodb/mongo/tree/r6.1.0-alpha is from April so it's probably too old. I bet various CSFLE features would not work as expected for example. Let's go with the alternative plan for now.

PR #230, opened 2022-09-12
test(NODE-4234): add aws-auth config for replica set

durran:


ShaneHarvey:
Does `"setParameter": {..."authenticationMechanisms": "MONGODB-AWS,..."` do anything? Drivers are already testing MONGODB-AWS without it IIUC.

durran:
> Does `"setParameter": {..."authenticationMechanisms": "MONGODB-AWS,..."` do anything? Drivers are already testing MONGODB-AWS without it IIUC.

I'll try without it as well. Won't pull this out of draft until I get it working with my Node PR.

ShaneHarvey:
I'm mistaken. I didn't realize that we already test MONGODB-AWS using the standalone variant of [auth-aws.json](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/orchestration/configs/servers/auth-aws.json). What's the motivation for adding a replica set variant?

PR #236, opened 2022-10-07
Add hygienic alternatives to Python virtual environment scripts

eramongodb:
## Description

This PR is the first of several PRs whose goal is to improve the robustness and hygiene of DET shell scripts by taking full advantage of the features provided by Bash.

This PR is focused on providing improved alternatives to scripts that pertain to Python virtual environment creation and activation. This PR does _not_ change the behavior of any existing scripts. Followup PRs will attempt to incorporate these new scripts into existing scripts, which may change their behavior.

## find-python3.sh

This file defines four functions:

- is_python3
- is_venv_capable
- is_virtualenv_capable
- find_python3

The primary purpose of this file is to provide `find_python3`, which is implemented in terms of the other three functions.

`find_python3` aims to serve as a reliable and comprehensive utility function to replace the numerous instances of "find a python3 binary capable of creating a virtual environment" conditional logic found in many Evergreen scripts, both in the DET repository (e.g. [activate_venv.sh](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/2443440ca69bcdba825eb0b43f874006ae3b5bca/.evergreen/csfle/activate_venv.sh#L1)) and in many other Drivers repositories. All demonstrate discrepancies from one another and are often not comprehensive in their handling of various variant-specific idiosyncracies.

Most notably, `find_python3` and other functions provided by `find-python3.sh` ensure no variables used within the functions are leaked into the parent shell. The name of the resulting python3 binary (and _only_ the name) is printed to `stdout` on success; all diagnostic and error messages are printed to `stderr` and may be silenced by redirecting `2>/dev/null`. This allows for convenient patterns such as the following:

```bash
$ cd ./.evergreen/csfle
$ . ../find-python3.sh
$ PYTHON="$(find_python3)" ./activate_venv.sh
```

This is unlike, for example, the current [utils.sh](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/2443440ca69bcdba825eb0b43f874006ae3b5bca/.evergreen/utils.sh#L6), which "leaks" the `$PYTHON` variable upon invocation of `venvcreate` and can inadvertently affect the behavior of other scripts if one is not careful, such as when subsequently using [set-temp-creds.sh](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/2443440ca69bcdba825eb0b43f874006ae3b5bca/.evergreen/csfle/set-temp-creds.sh#L36):

```bash
$ cd ./.evergreen/csfle
$ . ./activate-venv.sh  # Sets `PYTHON=/path/to/python` within venvcreate; installs boto3 package to venv.
$ . ./set-temp-creds.sh # Uses PYTHON set by venvcreate instead of virtual environment's python!
                        # boto3 package was installed using venv, not the system!
...
ModuleNotFoundError: No module named 'boto3'
```

This error can be worked-around by explicitly setting `PYTHON=python` after `activate_venv.sh` to ensure the venv python is used instead of any prior value of `$PYTHON`. However, it is preferable if such variable leaks did not occur in the first place.

## venv-utils.sh

This file is a hygienic alternative to [utils.sh](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/2443440ca69bcdba825eb0b43f874006ae3b5bca/.evergreen/utils.sh).

As described above, the `$PYTHON` variable (among others) is leaked by the `venvcreate` and `venvactivate` functions defined by `utils.sh`. Furthermore, their behavior is affected by the `$OS` environment variable parameter, which is often set and used by other Evergreen scripts in manners inconsistent to what `utils.sh` expects. This can have the effect of indirectly breaking the behavior of `utils.sh` functions. Lastly, `venvcreate` is defined to `exit 1` on error (terminates the _shell_), which limits error handling options by the user.

The `venvcreate` and `venvactivate` functions provided by `venv-utils.sh` resolves each of these issues. All parameters to each function is explicitly documented and asserted. The function's behavior depends solely on the provided arguments with no dependency on environment variables. On error, each function _immediately_ returns a non-zero exit code to allow callers to handle errors as appropriate to their needs.

## activate-kmstlsvenv.sh

This file is a hygienic alternative to [activate_venv.sh](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/2443440ca69bcdba825eb0b43f874006ae3b5bca/.evergreen/csfle/activate_venv.sh).

As with `activate_venv.sh`, this script is meant to be invoked by the caller to create and activate the `kmstlsvenv` virtual environment. It is implemented in terms of both `find-python3.sh` and `venv-utils.sh` to avoid the variable leakage problems described above. Furthermore, on error, it immediately returns a non-zero exit code instead of continuing execution as is currently done by `activate_venv.sh`.

## ShellCheck

All three files were analyzed by [ShellCheck](https://www.shellcheck.net/) using the command `shellcheck -x ./path/to/file.sh` from the root directory of the DET repository. Automated ShellCheck integration is outside the scope of this PR, but hopefully in the near future more DET scripts will be analyzed by ShellCheck to improve their quality.

### DIFF
--- a/.evergreen/csfle/activate-kmstlsvenv.sh
+++ b/.evergreen/csfle/activate-kmstlsvenv.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+#
+# activate-kmstlsvenv.sh
+#
+# Usage:
+#   . ./activate-kmstlsvenv.sh
+#
+# This file creates and activates the kmstlsvenv virtual environment in the
+# current working directory. This file must be invoked from within the
+# .evergreen/csfle directory in the Drivers Evergreen Tools repository.
+
+# Automatically invoked by activate-kmstlsvenv.sh.
+activate_kmstlsvenv() {
+  # shellcheck source=.evergreen/venv-utils.sh
+  . ../venv-utils.sh || return 1
+
+  if [[ -d kmstlsvenv ]]; then
+    venvactivate kmstlsvenv
+  else
+    # shellcheck source=.evergreen/find-python3.sh
+    . ../find-python3.sh || return 1
+
+    venvcreate "$(find_python3)" kmstlsvenv || return 1
+
+    CRYPTOGRAPHY_DONT_BUILD_RUST=1 python -m pip install --upgrade boto3~=1.19 cryptography~=3.4.8 pykmip~=0.10.0
+  fi
+}
+
+activate_kmstlsvenv
--- a/.evergreen/csfle/activate-kmstlsvenv.sh
+++ b/.evergreen/csfle/activate-kmstlsvenv.sh
@@ -12,15 +12,15 @@
 # Automatically invoked by activate-kmstlsvenv.sh.
 activate_kmstlsvenv() {
   # shellcheck source=.evergreen/venv-utils.sh
-  . ../venv-utils.sh || return 1
+  . ../venv-utils.sh || return
 
   if [[ -d kmstlsvenv ]]; then
     venvactivate kmstlsvenv
   else
     # shellcheck source=.evergreen/find-python3.sh
-    . ../find-python3.sh || return 1
+    . ../find-python3.sh || return
 
-    venvcreate "$(find_python3)" kmstlsvenv || return 1
+    venvcreate "$(find_python3)" kmstlsvenv || return
 
     CRYPTOGRAPHY_DONT_BUILD_RUST=1 python -m pip install --upgrade boto3~=1.19 cryptography~=3.4.8 pykmip~=0.10.0
   fi
--- a/.evergreen/find-python3.sh
+++ b/.evergreen/find-python3.sh
@@ -184,8 +184,14 @@ find_python3() (
     append_bins "/opt/mongodbtoolchain" "v[0-9]*" "bin/python3" "bin/python"
   } 1>&2
 
-  # For diagnostic purposes.
   {
+    # Some environments trigger an unbound variable error if "${bins[@]}" is empty when used below.
+    if (("${#bins[@]}" == 0)); then
+      echo "Could not find any python3 binaries!"
+      return 1
+    fi
+
+    # For diagnostic purposes.
     echo "List of python3 binaries to test:"
     for bin in "${bins[@]}"; do
       echo " - $bin"
--- a/.evergreen/find-python3.sh
+++ b/.evergreen/find-python3.sh
@@ -35,14 +35,10 @@ is_python3() (
   local -r bin="${1:?'is_python3 requires a name or path of a python binary to test'}"
 
   # Binary must be executable.
-  if ! command -V "$bin" &>/dev/null; then
-    return 1
-  fi
+  command -V "$bin" &>/dev/null || return 1
 
   # Reject binaries that do not support argument "-V".
-  if ! "$bin" -V &>/dev/null; then
-    return 1
-  fi
+  "$bin" -V &>/dev/null || return 1
 
   # Expect an output of the form: "Python x.y.z".
   # Note: Python 2 binaries output to stderr rather than stdout.
--- a/.evergreen/find-python3.sh
+++ b/.evergreen/find-python3.sh
@@ -44,12 +44,15 @@ is_python3() (
   # Note: Python 2 binaries output to stderr rather than stdout.
   local -r version_output="$("$bin" -V 2>&1)"
 
+  # For diagnostic purposes.
+  echo " - $bin: $version_output"
+
   # "Python x.y.z" -> "x.y.z"
   local -r version_str="$(perl -lne 'print $1 if m/.*([0-9]+\.[0-9]+\.[0-9]*)/' -- <(printf "%s" "$version_output"))"
 
   # Evaluate 3.0.0 <= x.y.z.
   sort -CV -- <(printf "%s\n%s\n" "3.0.0" "$version_str")
-)
+) 1>&2
 
 # is_venv_capable
 #
@@ -152,6 +155,8 @@ find_python3() (
   # The list of Python binaries to test for venv or virtualenv support.
   # The binaries are tested in the order of their position in the array.
   {
+    echo "Finding python3 binaries to test..."
+
     append_bins() {
       local -r path="${1:?'missing path'}"
       shift
--- a/.evergreen/find-python3.sh
+++ b/.evergreen/find-python3.sh
@@ -66,6 +66,7 @@ is_python3() (
 # Return 0 (true) if the given argument "$1" can successfully evaluate the
 # command:
 #   "$1" -m venv venv
+# and activate the created virtual environment.
 # Return a non-zero value (false) otherwise.
 #
 # Diagnostic messages may be printed to stderr (pipe 2). Redirect to /dev/null
@@ -81,8 +82,16 @@ is_venv_capable() (
   local -r tmp="$(mktemp -d)"
   trap 'rm -rf "$tmp"' EXIT
 
-  # Evaluate the result of this function.
-  "$bin" -m venv "$tmp" 1>&2
+  "$bin" -m venv "$tmp" || return 1
+
+  if [[ -f "$tmp/bin/activate" ]]; then
+    # shellcheck source=/dev/null
+    . "$tmp/bin/activate"
+  else
+    dos2unix "$tmp/Scripts/activate" || true
+    # shellcheck source=/dev/null
+    . "$tmp/Scripts/activate"
+  fi
 ) 1>&2
 
 # is_virtualenv_capable
@@ -97,6 +106,7 @@ is_venv_capable() (
 # Return 0 (true) if the given argument $1 can successfully evaluate the
 # command:
 #   "$1" -m virtualenv -p "$1" venv
+# and activate the created virtual environment.
 # Return a non-zero value (false) otherwise.
 #
 # Diagnostic messages may be printed to stderr (pipe 2). Redirect to /dev/null
@@ -112,8 +122,16 @@ is_virtualenv_capable() (
   local -r tmp="$(mktemp -d)"
   trap 'rm -rf "$tmp"' EXIT
 
-  # Evaluate the result of this function.
-  "$bin" -m virtualenv -p "$bin" "$tmp" 1>&2
+  "$bin" -m virtualenv -p "$bin" "$tmp" || return 1
+
+  if [[ -f "$tmp/bin/activate" ]]; then
+    # shellcheck source=/dev/null
+    . "$tmp/bin/activate"
+  else
+    dos2unix "$tmp/Scripts/activate" || true
+    # shellcheck source=/dev/null
+    . "$tmp/Scripts/activate"
+  fi
 ) 1>&2
 
 # find_python3
--- a/.evergreen/find-python3.sh
+++ b/.evergreen/find-python3.sh
@@ -1,9 +1,9 @@
 #!/usr/bin/env bash
 #
-# find-python.sh
+# find-python3.sh
 #
 # Usage:
-#   . /path/to/find-python.sh
+#   . /path/to/find-python3.sh
 #
 # This file defines the following utility functions:
 #   - is_python3
@@ -78,7 +78,7 @@ is_venv_capable() (
 
   local -r bin="${1:?'is_venv_capable requires a name or path of a python binary to test'}"
 
-  # Use a temporary directory to avoid polluting the caller's enviornment.
+  # Use a temporary directory to avoid polluting the caller's environment.
   local -r tmp="$(mktemp -d)"
   trap 'rm -rf "$tmp"' EXIT
 
@@ -109,7 +109,7 @@ is_virtualenv_capable() (
 
   local -r bin="${1:?'is_virtualenv_capable requires a name or path of a python binary to test'}"
 
-  # Use a temporary directory to avoid polluting the caller's enviornment.
+  # Use a temporary directory to avoid polluting the caller's environment.
   local -r tmp="$(mktemp -d)"
   trap 'rm -rf "$tmp"' EXIT
 
--- a/.evergreen/find-python3.sh
+++ b/.evergreen/find-python3.sh
@@ -47,7 +47,9 @@ is_python3() (
   # For diagnostic purposes.
   echo " - $bin: $version_output"
 
-  # Evaluate result of this function (zero == true).
+  # Evaluate result of this function.
+  # Note: Python True (1) and False (0) is treated as fail (1) and success (0)
+  # by Bash; therefore `is_python3` returns "true" when `v < 3` is false.
   "$bin" -c "import sys; exit(sys.version_info[0] < 3)"
 ) 1>&2
 
--- a/.evergreen/find-python3.sh
+++ b/.evergreen/find-python3.sh
@@ -47,11 +47,8 @@ is_python3() (
   # For diagnostic purposes.
   echo " - $bin: $version_output"
 
-  # "Python x.y.z" -> "x.y.z"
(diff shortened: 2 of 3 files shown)


ShaneHarvey:
I would prefer that we fix the seemingly small bugs in the current scripts rather than writing everything from scratch. Would these problems be solved by changing venvcreate/venvactivate to use private variables and changing set-temp-creds.sh to use "python" instead of "$PYTHON"?

eramongodb:
@ShaneHarvey I considered replacement as improvement, but due to the nature of the issues being addressed, I am concerned that doing so may have an unbounded scope of potentially breaking changes to existing users of these scripts. As the version of DET scripts being used by Drivers are often not pinned to a given commit (just a `git pull`), this could have the undesirable effect of unexpectedly and unconditionally breaking Drivers' Evergreen builds that may be deliberately or unknowningly depending on the current behavior of these scripts (leak of environment variables, use of environment variables, error handling and shell termination behavior, etc.). Thus the emphasis on, "This PR does _not_ change the behavior of any existing scripts".

By providing these scripts separately, Drivers will be given the opportunity to gradually test and migrate to using these hygienic scripts before further behavior-changing modifications are introduced to DET scripts (I am eyeing the `start-orchestration.sh` script in particular). Eventually, the current scripts can then be deprecated and removed in a controlled manner once these new alternatives have been sufficiently battle-tested and adopted.

ShaneHarvey:
If this was a change that we knew would break all driver testing I might agree with this approach but I don't think that's the position we're in. The minor tweaks to avoid shadowing or unintentionally reusing $PYTHON seem like small fixes that have low chance of breaking anyone. The plus side of my suggestion is that in the best case we fix these bugs and drivers don't need to change anything at all.

The normal approach we'd use is to fix the bug, test it in 1 or 2 drivers, merge, and then send an announcement in #drivers that teams _might_ see python env setup failures in CSFLE/MONGODB-AWS test suites with a suggestion of how to fix it.

.evergreen/find-python3.sh
```diff
@@ -0,0 +1,232 @@
...
+# Diagnostic messages may be printed to stderr (pipe 2). Redirect to /dev/null
+# to silence these messages.
+is_venv_capable() (
+  set -o errexit
+  set -o nounset
+  set -o pipefail
+
+  local -r bin="${1:?'is_venv_capable requires a name or path of a python binary to test'}"
+
+  # Use a temporary directory to avoid polluting the caller's enviornment.
```
rcsanchez97, line 81:
```suggestion
  # Use a temporary directory to avoid polluting the caller's environment.
```

.evergreen/find-python3.sh
```diff
@@ -0,0 +1,232 @@
...
+# Diagnostic messages may be printed to stderr (pipe 2). Redirect to /dev/null
+# to silence these messages.
+is_virtualenv_capable() (
+  set -o errexit
+  set -o nounset
+  set -o pipefail
+
+  local -r bin="${1:?'is_virtualenv_capable requires a name or path of a python binary to test'}"
+
+  # Use a temporary directory to avoid polluting the caller's enviornment.
```
rcsanchez97, line 112:
```suggestion
  # Use a temporary directory to avoid polluting the caller's environment.
```

.evergreen/find-python3.sh
```diff
@@ -0,0 +1,232 @@
...
+        for bin in "${suffixes[@]}"; do
+          if is_python3 "$dir/$bin"; then
+            bins+=("$dir/$bin")
+          fi
+        done
+      done
+    }
+
+    # /opt/mongodbtoolchain/vX/bin/python
+    append_bins "/opt/mongodbtoolchain" "v[0-9]*" "bin/python3" "bin/python"
```
rcsanchez97, line 176: This should be reconsidered.  We use the toolchain in various places and those uses have the potential to become problematic.  Specifically, the toolchain is intended for the server project, so it can break without warning.  It may be that the way this script checks for presence and capability of Python is resilient against unexpected changes to the toolchain.  If that's the case then a comment to that effect might be good here.

rcsanchez97 (commented):
I can understand the concerns of Shane, though I don't share them.  It seems like a step in the right direction to make our builds more hygienic and robust and this approach is superior to incremental bug fixes to the existing scripts.  That said, I'll wait on approving to see what others might have to say.

benjirewis (approved):
These scripts LGTM; cool bash coding, thanks @eramongodb 🧑‍🔧 . 

I do wonder if we could just replace the old scripts with these new scripts as part of this PR, though. Doing it in separate PRs just seems like prolonging the possible breaking of drivers' reliance on the scripts.

kevinAlbs (approved):
Changes LGTM.

> If this was a change that we knew would break all driver testing I might agree with this approach but I don't think that's the position we're in. The minor tweaks to avoid shadowing or unintentionally reusing $PYTHON seem like small fixes that have low chance of breaking anyone.

[Testing the `activate-kmstlsvenv.sh`](https://spruce.mongodb.com/version/63445333c9ec4468fb466795/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) script with Go suggests there are required changes to update Evergreen config to explicitly use bash:

```sh
[2022/10/10 17:17:53.902] sh: 17: ./activate-kmstlsvenv.sh: [[: not found
[2022/10/10 17:17:53.902] sh: 52: ../find-python3.sh: Syntax error: "(" unexpected
[2022/10/10 17:17:53.903] Command ''shell.exec' in "start-cse-servers"' failed: shell script encountered problem: exit code 2.
```

I suggest filing a DRIVERS ticket to request drivers use the new scripts. Once all drivers have implemented the required changes, the existing scripts can be safely updated.

.evergreen/find-python3.sh
```diff
@@ -0,0 +1,232 @@
+#!/usr/bin/env bash
+#
+# find-python.sh
```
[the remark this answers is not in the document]
eramongodb: Fixed.

.evergreen/find-python3.sh
```diff
@@ -0,0 +1,232 @@
...
+#
+# Usage:
+#   . /path/to/find-python.sh
```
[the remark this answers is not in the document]
eramongodb: Fixed.

ShaneHarvey:
> [Testing the activate-kmstlsvenv.sh](https://spruce.mongodb.com/version/63445333c9ec4468fb466795/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) script with Go suggests there are required changes to update Evergreen config to explicitly use bash:

This was the point I was trying to make. I don't think we should make all teams update their scripts to use this new approach. Instead we can make the small $PYTHON bug fixes to the existing scripts and teams (most likely) wouldn't need to do any work.

I do like the modern bash and nice error messages though.

kevinAlbs:
> This was the point I was trying to make. I don't think we should make all teams update their scripts to use this new approach. Instead we can make the small $PYTHON bug fixes to the existing scripts and teams (most likely) wouldn't need to do any work.
> 
> I do like the modern bash and nice error messages though.

I see. I misinterpreted the point. I am not opposed to the small fixes in the existing scripts. But IMO this PR is still an overall improvement. I do not consider it a requirement to update the existing scripts to merge.

eramongodb:
> we can make the small $PYTHON bug fixes to the existing scripts and teams (most likely) wouldn't need to do any work

This may be somewhat feasible for `activate-kmstlsvenv.sh`, but not for `venv-utils.sh`.

These scripts were motivated by external discussions suggesting that Evergreen supports Bash on all distros, even on Windows via Cygwin (see also the `shell` parameter for the [shell.exe Evergreen project command](https://github.com/evergreen-ci/evergreen/wiki/Project-Commands#shell-exec)). This guarantee is what permitted me to make full use of Bash features in order to accomplish the goal(s) of this PR.

Running ShellCheck on `activate-kmstlsvenv.sh` with the shebang set to invoke `sh` yields the following:

```
In .evergreen/csfle/activate-kmstlsvenv.sh line 17:
  if [[ -d kmstlsvenv ]]; then
     ^-----------------^ SC2039: In POSIX sh, [[ ]] is undefined.

For more information:
  https://www.shellcheck.net/wiki/SC2039 -- In POSIX sh, [[ ]] is undefined.
```

Similarly, running ShellCheck on `venv-utils.sh` with the shebang set to invoke `sh` yields the following (repeats of warnings omitted for brevity):

```
In .evergreen/venv-utils.sh line 27:
  local -r bin="${1:?'venvcreate requires a Python binary to use for the virtual environment'}"
  ^----------^ SC2039: In POSIX sh, 'local' is undefined.


In .evergreen/venv-utils.sh line 30:
  if "$bin" -m venv -h &>/dev/null; then
                       ^---------^ SC2039: In POSIX sh, &> is undefined.


In .evergreen/venv-utils.sh line 44:
  if [[ -f "$path/Scripts/activate" ]]; then
     ^-- SC2039: In POSIX sh, [[ ]] is undefined.
```

The `[[ ]]` warnings can be easily addressed by demoting the conditional construct used to the Bourne shell builtin `[ ]` and `test` (the Bash builtin `[[ ]]` was used in these scripts as the [new, improved version](https://mywiki.wooledge.org/BashFAQ/031) of `[ ]` and `test`). Similarly, the `&>` warnings can be addressed by using the equivalent `1>/dev/null 2>&1` instead of `&>/dev/null`.

However, the `local` warnings are not addressable, as POSIX `sh` [does not support](https://www.austingroupbugs.net/bug_view_page.php?bug_id=767) `local` scoping of variables. None of the [workarounds](https://stackoverflow.com/a/18600920) seem applicable (subshells prevent use of functions as `source`) or recommendable (complex boilerplate, not scalable).

Given the primary goal of this PR is to avoid variable leakage, being unable to use Bash in full due to semi-backwards-compatibility requirements with current use of `activate_venv.sh` and `utils.sh` (as observed when testing with Go) would significantly reduce the value of this PR.

.evergreen/csfle/activate-kmstlsvenv.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
```
ShaneHarvey, line 1: Aren't shebangs meaningless (ignored) for scripts that are intended to be sourced? Eg changing this to `#!/usr/bin/env sh` would have no effect.

.evergreen/find-python3.sh
```diff
@@ -0,0 +1,234 @@
...
+  set -o pipefail
+
+  local -r bin="${1:?'is_venv_capable requires a name or path of a python binary to test'}"
+
+  # Use a temporary directory to avoid polluting the caller's environment.
+  local -r tmp="$(mktemp -d)"
+  trap 'rm -rf "$tmp"' EXIT
+
+  # Evaluate the result of this function.
+  "$bin" -m venv "$tmp" 1>&2
```
ShaneHarvey, line 82: I recall that in some broken environments python -m venv/virtualenv can succeed but the activate script fails. But no need to change anything now. Let's wait until we actually encounter such a failure before we try to workaround it.

ShaneHarvey (commented):
My understanding is that we don't _need_ to use bash local variables, for example we can prefix the vars to avoid clashing in practice:
```
venvcreate () {
+    __PYTHON="$1"
-    PYTHON="$1"
```
Another suitable option would be to remove the var altogether and use "$1" instead of "$PYTHON".

eramongodb (commented):
> we can prefix the vars to avoid clashing in practice

This is not a proper solution, as it does not address the issue of variable leakage. This just changes the name of the variable being leaked. The environment will still be polluted by variables. This is not scalable.

> Another suitable option would be to remove the var altogether and use "$1" instead of "$PYTHON".

This is also not a scalable solution. Forbidding the use of named variables in non-subshell functions in order to continue supporting the Bourne shell (where `local` is not available) does not align with the goals of this PR, which as stated earlier, "is to improve the robustness and hygiene of DET shell scripts by taking full advantage of the features provided by Bash".

.evergreen/csfle/activate-kmstlsvenv.sh
ShaneHarvey: Gotcha thanks for explaining that.

ShaneHarvey (approved):

eramongodb:
Latest changes are verified by [this patch](https://spruce.mongodb.com/version/63482aa33066150f1e093af6). The set of distros tested are (best-effort) all non-EOL'd distros as currently listed in the MongoDB Platform Roadmap. The [additions to the Evergreen config](https://evergreen.mongodb.com/filediff/63482aa33066150f1e093af6/?file_name=.evergreen%2Fconfig.yml&patch_number=0) are omitted from this PR, but may be included if desirable.

The root cause of the [test failure on macos-1012](https://spruce.mongodb.com/task/drivers_tools_test_find_python3_macos_1012_test_activate_kmstlsvenv_patch_2443440ca69bcdba825eb0b43f874006ae3b5bca_63482aa33066150f1e093af6_22_10_13_15_11_32/logs?execution=0) is being tracked by [BUILD-16106](https://jira.mongodb.org/browse/BUILD-16106).

.evergreen/find-python3.sh
```diff
@@ -0,0 +1,257 @@
...
+  local -r version_output="$("$bin" -V 2>&1)"
+
+  # For diagnostic purposes.
+  echo " - $bin: $version_output"
+
+  # "Python x.y.z" -> "x.y.z"
+  local -r version_str="$(perl -lne 'print $1 if m/.*([0-9]+\.[0-9]+\.[0-9]*)/' -- <(printf "%s" "$version_output"))"
+
+  # Evaluate 3.0.0 <= x.y.z.
+  sort -CV -- <(printf "%s\n%s\n" "3.0.0" "$version_str")
```
ShaneHarvey, line 54: FYI this logic can be implemented much simpler in python. For example:
```
if python -c "import sys; exit(sys.version_info[0] < 3)"; then
   echo "python 3"
else
   echo "python 2"
fi
```

this would avoid the dependency on "sort".

eramongodb: [Verified](https://spruce.mongodb.com/version/63485ef132f417225c2244fb/tasks) this suggestion works as suggested. Thank you!

However, it does not completely remove the GNU Coreutils `sort` dependency as it is used below to prioritize newer MongoDB Toolchain and Python versions under `/opt`. It seems like macos-1012 is [circumstantially passing](https://evergreen.mongodb.com/lobster/evergreen/task/drivers_tools_test_find_python3_macos_1012_test_activate_kmstlsvenv_patch_2443440ca69bcdba825eb0b43f874006ae3b5bca_63485ef132f417225c2244fb_22_10_13_18_54_41/0/task#bookmarks=0%2C406&l=1&shareLine=245) due to system `python3` satisfying the requirements despite `sort` command failures when searching for binaries under `/opt`.

eramongodb:
Unless there are any remaining concerns, I think I am comfortable merging this PR given [all distros currently supported](https://spruce.mongodb.com/version/63485ef132f417225c2244fb) according to the MongoDB Platform Roadmap are currently passing despite [BUILD-16106](https://jira.mongodb.org/browse/BUILD-16106).

The plan is to then create a DRIVERS Task instructing Drivers to replace use of `activate_venv.sh` and `utils.sh` with `csfle/activate-kmstlsvenv.sh` and `venv_utils.sh` respectively. Once all Drivers have completed this task, the old `activate_venv.sh` and `utils.sh` scripts can be safely removed from this repo.

eramongodb:
> The plan is to then create a DRIVERS Task instructing Drivers to replace use of `activate_venv.sh` and `utils.sh` with `csfle/activate-kmstlsvenv.sh` and `venv_utils.sh` respectively.

[DRIVERS-2497](https://jira.mongodb.org/browse/DRIVERS-2497) has been created.

PR #240, opened 2022-10-24
PYTHON-3487 Prefer venv over global virtualenv

blink1073:


blink1073:
All tests passed in EG

PR #237, opened 2022-10-12
Update server versions in download-mongodb.sh

kevinAlbs:
# Summary

- update VERSION_RAPID to 6.1.0
- update patch versions of server downloads
- remove MongoDB server 6.0 from Debian 9.2

# Background & Motivation

Release versions were copied from https://www.mongodb.com/docs/manual/release-notes/

Here is a patch build using this branch in the Go driver: https://spruce.mongodb.com/version/6346f327a4cf4706ba3d51eb/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

PR #242, opened 2022-10-28
DRIVERS-2459: Add script to download mongocryptd and crypt_shared

alcaeus:
[DRIVERS-2459](https://jira.mongodb.org/browse/DRIVERS-2459)

[DRIVERS-2400](https://jira.mongodb.org/browse/2400) asks drivers to run CSFLE tests on Serverless, which require mongocryptd or crypt_shared to run. Since this is installed in run-orchestration.sh which isn't invoked on Serverless, this PR adds the necessary download steps to create-instance.sh. A new environment variable allows disabling this download in case people want to test this locally (where they typically would have mongocryptd or crypt_shared installed already).

For now the script always downloads the rapid release version - this is to avoid parsing the reported Serverless version and deducing the correct version string to pass along (e.g. `6.1.3` would have to be replaced with `rapid`). This can be added in a separate step if deemed necessary.

* [Patch build](https://spruce.mongodb.com/version/635b93b39ccd4e5dcf2654f6/tasks)
* [PHPLIB changes](https://github.com/mongodb/mongo-php-library/commit/c0ae6cfc9f4f990b93949fa181e9517e917ee661) to remove manual download and use this PR version of drivers-evergreen-tools

### DIFF
--- a/.evergreen/download-mongodb.sh
+++ b/.evergreen/download-mongodb.sh
@@ -494,8 +494,7 @@ get_mongodb_download_url_for ()
          if [ -z $MONGODB_60 ]; then
            VERSION_INCLUDES_CRYPT_SHARED=NO
          fi ;;
-      rapid) MONGODB_DOWNLOAD_URL=$MONGODB_RAPID
-         VERSION_INCLUDES_CRYPT_SHARED=NO ;;
+      rapid) MONGODB_DOWNLOAD_URL=$MONGODB_RAPID ;;
       v6.0-latest) MONGODB_DOWNLOAD_URL=$MONGODB_60_LATEST ;;
       6.0) MONGODB_DOWNLOAD_URL=$MONGODB_60 ;;
       5.0) MONGODB_DOWNLOAD_URL=$MONGODB_50
--- a/.evergreen/download-mongodb.sh
+++ b/.evergreen/download-mongodb.sh
@@ -520,7 +520,10 @@ get_mongodb_download_url_for ()
          VERSION_INCLUDES_CRYPT_SHARED=NO ;;
    esac
 
-   [ -z "$MONGODB_DOWNLOAD_URL" ] && MONGODB_DOWNLOAD_URL="Unknown version: $_VERSION for $_DISTRO"
+   if [ -z "$MONGODB_DOWNLOAD_URL" ]; then
+     echo "Unknown version: $_VERSION for $_DISTRO"
+     exit 1
+   fi
 
    if [ "$VERSION_INCLUDES_CRYPT_SHARED" = "YES" ]; then
       # The crypt_shared package is simply the same file URL with the "mongodb-"
--- a/.evergreen/serverless/create-instance.sh
+++ b/.evergreen/serverless/create-instance.sh
@@ -9,7 +9,7 @@ set +o xtrace # Disable xtrace to ensure credentials aren't leaked
 #   SERVERLESS_DRIVERS_GROUP    Required. Atlas group for drivers testing.
 #   SERVERLESS_API_PUBLIC_KEY   Required. Public key for Atlas API request.
 #   SERVERLESS_API_PRIVATE_KEY  Required. Private key for Atlas API request.
-#   SERVERLESS_SKIP_CRYPT       Optional. If set, skips installing mongocryptd and crypt_shared (defaults to true)
+#   SERVERLESS_SKIP_CRYPT       Optional. If set, skips installing mongocryptd and crypt_shared (defaults to "ON")
 #
 # On success, this script will output serverless-expansion.yml with the
 # following expansions:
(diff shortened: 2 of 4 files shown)


.evergreen/serverless/create-instance.sh
```diff
@@ -113,6 +114,12 @@ SINGLE_ATLASPROXY_SERVERLESS_URI: "$SERVERLESS_URI"
 MULTI_ATLASPROXY_SERVERLESS_URI: "$SERVERLESS_URI"
 SERVERLESS_MONGODB_VERSION: "$SERVERLESS_MONGODB_VERSION"
 EOF
+
+        if [ "$SERVERLESS_SKIP_CRYPT" != "OFF" ]; then
```
DmitryLukyanov, line 118: given a default value is `true`, should we use `false` here instead `OFF`?

.evergreen/serverless/download-crypt.sh
```diff
@@ -0,0 +1,29 @@
...
+# Functions to fetch MongoDB binaries
+. $DIR/../download-mongodb.sh
+
+get_distro
+get_mongodb_download_url_for "$DISTRO" "$MONGODB_VERSION"
+
+# Don't need to install the legacy shell
+download_and_extract_package "$MONGODB_DOWNLOAD_URL" "$EXTRACT"
+
+if [ -z $MONGO_CRYPT_SHARED_DOWNLOAD_URL ]; then
```
DmitryLukyanov, line 15: can we call this code here: https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/run-orchestration.sh#L105?

.evergreen/serverless/create-instance.sh
```diff
@@ -9,6 +9,7 @@ set +o xtrace # Disable xtrace to ensure credentials aren't leaked
 #   SERVERLESS_DRIVERS_GROUP    Required. Atlas group for drivers testing.
 #   SERVERLESS_API_PUBLIC_KEY   Required. Public key for Atlas API request.
 #   SERVERLESS_API_PRIVATE_KEY  Required. Private key for Atlas API request.
+#   SERVERLESS_SKIP_CRYPT       Optional. If set, skips installing mongocryptd and crypt_shared (defaults to true)
```
DmitryLukyanov, line 12: do you assign this anywhere?

.evergreen/serverless/create-instance.sh
alcaeus: I've updated the comment to note the default value as `ON`. I used `ON` and `OFF` for consistency with the `LOADBALANCED` variable.

.evergreen/serverless/create-instance.sh
alcaeus: The variable is only used in line 118. It is not used on Serverless and only there for people that want to run `create-instance.sh` locally where they already have mongodb and everything installed and want to skip downloading it again.

.evergreen/serverless/download-crypt.sh
alcaeus: I considered that but decided against it due to the refactoring needed to have the code shared between run-orchestration.sh and create-instance.sh. I can take a look at consolidating the logic separately.

DmitryLukyanov: >I can take a look at consolidating the logic separately.

that would be good

.evergreen/download-mongodb.sh
```diff
@@ -494,8 +494,7 @@ get_mongodb_download_url_for ()
          if [ -z $MONGODB_60 ]; then
            VERSION_INCLUDES_CRYPT_SHARED=NO
          fi ;;
-      rapid) MONGODB_DOWNLOAD_URL=$MONGODB_RAPID
-         VERSION_INCLUDES_CRYPT_SHARED=NO ;;
+      rapid) MONGODB_DOWNLOAD_URL=$MONGODB_RAPID ;;
```
alcaeus, line 497: I believe this was missed when `VERSION_RAPID` was updated to 6.1 in #237. The previous rapid version (5.3) didn't include crypt_shared, and I believe that's why this was included. Since the rapid release will now always include crypt_shared, I believe this should be safe to remove. Any objections to this @kevinAlbs?

DmitryLukyanov (approved):
LGTM, checked with [c#](https://evergreen.mongodb.com/task_log_raw/dot_net_driver_serverless_tests_ubuntu__os~ubuntu_1804_auth~auth_ssl~ssl_compressor~zlib_test_serverless_netstandard21_patch_ca00c14f882ff88234697a23f585c95648d9c031_635be905850e614b0ee0db52_22_10_28_14_36_54/1?type=T#L268)

.evergreen/download-mongodb.sh
kevinAlbs: SGTM. Thank you.

PR #244, opened 2022-11-10
DRIVERS-2497 Fix paths on Cygwin and Python package dependencies

eramongodb:
## Description

This PR is a followup to https://github.com/mongodb-labs/drivers-evergreen-tools/pull/236 and applies a critical patch to [DRIVERS-2497](https://jira.mongodb.org/browse/DRIVERS-2497) required by Windows distros.

This PR is verified by [this patch](https://spruce.mongodb.com/version/636d66b17742ae6eb49bfedf/tasks).

## Paths on Cygwin

The [tests](https://spruce.mongodb.com/version/63485ef132f417225c2244fb) used to validate behavior in https://github.com/mongodb-labs/drivers-evergreen-tools/pull/236 did not account for the presence of [pre commands](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/f5d1c049948718218b006c0732990e1c642cbb13/.evergreen/config.yml#L424) which modifies the environment such that it does not accurately reflect the environment used by Drivers, such as [modifying the `$PATH` variable](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/f5d1c049948718218b006c0732990e1c642cbb13/.evergreen/config.yml#L80) to prefer binaries provided by `$MONGODB_BINARIES` such as `mktemp`. This [hid Cygwin path conversions requirements](https://evergreen.mongodb.com/lobster/evergreen/task/drivers_tools_test_kmstlsvenv_windows_64_2019_test_kmstlsvenv_patch_f5d1c049948718218b006c0732990e1c642cbb13_636a8f8e306615468cab5d20_22_11_08_17_19_11/0/task#bookmarks=0%2C574&l=1&shareLine=302) by Python binaries on Windows when creating the virtual environment, demonstrated below on a windows-64-vs2017-large distro (`virtualenv` is used to obtain informative output; `venv` demonstrates similar behavior but without any output):

```bash
$ pwd
/home/Administrator

$ # As found by find_python3.
$ PYTHON="C:/python/Python310/python.exe"

$ # Command behaves as expected given a relative path.
$ # Note `dest=C:\cygwin\home\Administrator\venv` in output.
$ "$PYTHON" -m virtualenv -p "$PYTHON" venv
created virtual environment CPython3.10.2.final.0-64 in 5641ms
  creator CPython3Windows(dest=C:\cygwin\home\Administrator\venv, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=C:\Users\Administrator\AppData\Local\pypa\virtualenv)
    added seed packages: pip==22.2.2, setuptools==65.4.1, wheel==0.37.1
  activators BashActivator,BatchActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator

$ find venv -mindepth 1 -maxdepth 1
venv/.gitignore
venv/Lib
venv/pyvenv.cfg
venv/Scripts

$ # As done by is_venv_capable and is_virtualenv_capable.
$ # /tmp/tmp.RukOr5qrGb
$ VENV="$(mktemp -d)"

$ # Command succeeds but places virtual environment in an unxpected location.
$ # Note `dest=C:\tmp\tmp.RukOr5qrGb` in output.
$ "$PYTHON" -m virtualenv -p "$PYTHON" "$VENV"
created virtual environment CPython3.10.2.final.0-64 in 1047ms
  creator CPython3Windows(dest=C:\tmp\tmp.RukOr5qrGb, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=C:\Users\Administrator\AppData\Local\pypa\virtualenv)
    added seed packages: pip==22.2.2, setuptools==65.4.1, wheel==0.37.1
  activators BashActivator,BatchActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator

$ # Intended directory is unexpectedly empty despite successful command.
$ find "$VENV" -maxdepth 0 -empty
/tmp/tmp.RukOr5qrGb
```

A sanity check was added to the `is_venv_capable` and `is_virtualenv_capable` functions to ensure correct behavior, as well as an explicit check for the presence of an activation script to provide more informative error messages if one still cannot be found.

## Seed Packages

In addition to ensuring `venvcreate` handles paths correctly on Cygwin, the `venvcreate` function was updated to ensure all three "seed" packages `pip`, `setuptools`, and `wheel` are consistently installed in the virtual environment, as default behavior is inconsistent depending on `venv` vs. `virtualenv` and their respective versions.

A drive-by fix to correctly pass `-p "$bin"` when using the `virtualenv` module was also applied.

The `--no-cache-dir` argument was removed due to lack of necessity.

The `--system-site-packages` argument was added to improve script performance.

Error handling of the `venvcreate` function was improved to ensure the virtual environment is only activated on success. `deactivate` is only possible/necessary if `venvcreate` was successful.

## kmstlsvenv Packages

As a result of ensuring up-to-date `pip`, `setuptools`, and `wheel` packages in the virtual environment, some distros began to encounter issues with installing required packages for the kmstlsvenv virtual environment. I took this opportunity to strictly narrow down the scope and conditions when default behavior does not suffice for successful installation.

The actual packages required by kmstlsvenv scripts, `boto3` and `pykmip`, are still pinned to `~=1.19.0` and `~=0.10.0` respectively. All additional, conditionally pinned packages are dependencies required by these two packages.

The `greenlet` package is conditionally pinned to `<2.0` to avoid build failures on macos-1012.

The `setuptools` package is conditionally pinned to `<65.0` to avoid build failures on windows-64-2016 (see [BUILD-16233](https://jira.mongodb.org/browse/BUILD-16233)).

The `cryptography` package is conditionally pinned to `<3.4` to avoid dependency on the presence of a Rust compiler when a cryptography wheel is not available. The associated conditions were narrowed down as much as possible to allow/encourage use of up-to-date packages whenever possible.

As with `venvcreate`, error handling of the `activate_kmstlsvenv` function was improved to ensure the virtual environment is only activated on success. `deactivate` is only possible/necessary if `activate_kmstlsvenv` was successful.

### DIFF
--- a/.evergreen/csfle/activate-kmstlsvenv.sh
+++ b/.evergreen/csfle/activate-kmstlsvenv.sh
@@ -5,9 +5,17 @@
 # Usage:
 #   . ./activate-kmstlsvenv.sh
 #
-# This file creates and activates the kmstlsvenv virtual environment in the
+# This file creates and/or activates the kmstlsvenv virtual environment in the
 # current working directory. This file must be invoked from within the
 # .evergreen/csfle directory in the Drivers Evergreen Tools repository.
+#
+# If a kmstlsvenv virtual environment already exists, it will be activated and
+# no further action will be taken. If a kmstlsvenv virtual environment must be
+# created, required packages will also be installed.
+
+# If an error occurs during creation, activation, or installation of packages,
+# the kmstlsvenv virtual environment will be deactivated and activate_kmstlsvenv
+# will return a non-zero value.
 
 # Automatically invoked by activate-kmstlsvenv.sh.
 activate_kmstlsvenv() {
@@ -15,14 +23,75 @@ activate_kmstlsvenv() {
   . ../venv-utils.sh || return
 
   if [[ -d kmstlsvenv ]]; then
-    venvactivate kmstlsvenv
+    venvactivate kmstlsvenv || return
   else
     # shellcheck source=.evergreen/find-python3.sh
     . ../find-python3.sh || return
 
     venvcreate "$(find_python3)" kmstlsvenv || return
 
-    CRYPTOGRAPHY_DONT_BUILD_RUST=1 python -m pip install --upgrade boto3~=1.19 cryptography~=3.4.8 pykmip~=0.10.0
+    local packages=(
+      "boto3~=1.19.0"
+      "pykmip~=0.10.0"
+    )
+
+    if [[ "$OSTYPE" == darwin16 && "$HOSTTYPE" == x86_64 ]]; then
+      # Avoid `error: thread-local storage is not supported for the current
+      # target` on macos-1012.
+      packages+=("greenlet<2.0")
+    fi
+
+    if [[ "$OSTYPE" == cygwin && "$HOSTTYPE" == x86_64 ]]; then
+      local -r is_win_2016="$(systeminfo.exe /FO LIST | perl -lne 'print $1 if m/^OS Name:\s+(.*)$/' || true)"
+
+      if [[ "$is_win_2016" =~ 2016 ]]; then
+        # Avoid `RuntimeError: Could not determine home directory.` on
+        # windows-64-2016. See BUILD-16233.
+        python -m pip install -U "setuptools<65.0" || {
+          local -r ret="$?"
+          deactivate || return 1 # Deactivation should never fail!
+          return "$ret"
+        }
+      fi
+    fi
+
+    # Avoid `error: can't find Rust compiler`.
+    if [[ "$OSTYPE" =~ linux && ! -f /etc/os-release ]]; then
+      # rhel62-* is the only suppoerted Linux-like distro that does not provide
+      # /etc/os-release. Remove this condition once support for rhel62 is
+      # dropped.
+      packages+=("cryptography<3.4")
+    elif [[ "$OSTYPE" =~ linux ]]; then
+      local -r os_id="$(perl -lne 'print $1 if m/^ID="?([^"]+)"?/' /etc/os-release || true)"
+      local -r os_ver="$(perl -lne 'print $1 if m/^VERSION_ID="?([^"]+)"?/' /etc/os-release || true)"
+
+      case "$os_id" in
+      rhel)
+        if [[ "$HOSTTYPE" =~ (powerpc64le|s390x) ]]; then
+          # rhelXY-power8-* and rhelXY-zseries-*
+          packages+=("cryptography<3.4")
+        fi
+        ;;
+      sles)
+        if [[ "$os_ver" == 12.3 && "$HOSTTYPE" == s390x ]]; then
+          # suse12-zseries-*
+          packages+=("cryptography<3.4")
+        fi
+        ;;
+      ubuntu)
+        if [[ "$os_ver" == 18.04 && "$HOSTTYPE" =~ (s390x|powerpc64le) ]]; then
+          # ubuntu1804-power8-* and ubuntu1804-zseries-*
+          packages+=("cryptography<3.4")
+        fi
+        ;;
+      esac
+    fi
+
+    python -m pip install -U "${packages[@]}" || {
+      local -r ret="$?"
+      deactivate || return 1 # Deactivation should never fail!
+      return "$ret"
+    }
   fi
 }
 
--- a/.evergreen/csfle/activate-kmstlsvenv.sh
+++ b/.evergreen/csfle/activate-kmstlsvenv.sh
@@ -57,7 +57,7 @@ activate_kmstlsvenv() {
 
     # Avoid `error: can't find Rust compiler`.
     if [[ "$OSTYPE" =~ linux && ! -f /etc/os-release ]]; then
-      # rhel62-* is the only suppoerted Linux-like distro that does not provide
+      # rhel62-* is the only supported Linux-like distro that does not provide
       # /etc/os-release. Remove this condition once support for rhel62 is
       # dropped.
       packages+=("cryptography<3.4")
--- a/.evergreen/csfle/activate-kmstlsvenv.sh
+++ b/.evergreen/csfle/activate-kmstlsvenv.sh
@@ -42,9 +42,9 @@ activate_kmstlsvenv() {
     fi
 
     if [[ "$OSTYPE" == cygwin && "$HOSTTYPE" == x86_64 ]]; then
-      local -r is_win_2016="$(systeminfo.exe /FO LIST | perl -lne 'print $1 if m/^OS Name:\s+(.*)$/' || true)"
+      local -r windows_os_name="$(systeminfo.exe /FO LIST | perl -lne 'print $1 if m/^OS Name:\s+(.*)$/' || true)"
 
-      if [[ "$is_win_2016" =~ 2016 ]]; then
+      if [[ "$windows_os_name" =~ 2016 ]]; then
         # Avoid `RuntimeError: Could not determine home directory.` on
         # windows-64-2016. See BUILD-16233.
         python -m pip install -U "setuptools<65.0" || {
--- a/.evergreen/csfle/activate-kmstlsvenv.sh
+++ b/.evergreen/csfle/activate-kmstlsvenv.sh
@@ -55,43 +55,24 @@ activate_kmstlsvenv() {
       fi
     fi
 
-    # Avoid `error: can't find Rust compiler`.
-    if [[ "$OSTYPE" =~ linux && ! -f /etc/os-release ]]; then
-      # rhel62-* is the only supported Linux-like distro that does not provide
-      # /etc/os-release. Remove this condition once support for rhel62 is
-      # dropped.
+    if ! python -m pip install -U "${packages[@]}"; then
+      # Avoid `error: can't find Rust compiler`.
+      # Assume install failure at this point is due to new versions of
+      # cryptography require a Rust toolchain when a cryptography wheel is not
+      # present due to the package versions available.  This is required by at
+      # least the following distros (by OS and host):
+      #  - RHEL 6.2
+      #  - All RHEL on powerpc64le or s390x.
+      #  - OpenSUSE 12 on s390x.
+      #  - Ubuntu 18.04 on powerpc64le or s390x
       packages+=("cryptography<3.4")
-    elif [[ "$OSTYPE" =~ linux ]]; then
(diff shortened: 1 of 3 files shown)


rcsanchez97 (approved):
LGTM

If my confusion is unfounded, then feel free to ignore it, otherwise consider clarifying the comment a bit.

.evergreen/csfle/activate-kmstlsvenv.sh
```diff
@@ -5,24 +5,93 @@
...
 # current working directory. This file must be invoked from within the
 # .evergreen/csfle directory in the Drivers Evergreen Tools repository.
+#
+# If a kmstlsvenv virtual environment already exists, it will be activated and
+# no further action will be taken. If a kmstlsvenv virtual environment must be
+# created, required packages will also be installed.
+
+# If an error occurs during creation, activation, or installation of packages,
+# the kmstlsvenv virtual environment will be deactivated and activate_kmstlsvenv
+# will return a non-zero value.
```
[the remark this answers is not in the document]
eramongodb: `return` without an argument returns the last exit code. It is akin to `throw;` in C++ (rethrow the caught exception):

```bash
$ example() { (exit 42) || return; }
$ example; echo $?
42
```

.evergreen/venv-utils.sh
```diff
@@ -21,38 +21,73 @@
...
-      # mongovenv/Scripts/activate: line 3: $'\r': command not found
-      if [[ -f "$path/Scripts/activate" ]]; then
-        dos2unix "$path/Scripts/activate" || return
-      fi
+    case "$mod" in
+    venv)
+      "$bin" -m "$mod" --system-site-packages "$real_path" || continue
+      ;;
+    virtualenv)
+      # -p: ensure correct Python binary is used by virtual environment.
```
ShaneHarvey, line 55: Is -p actually needed here? `-p` defaults to the current version of python so this seems redundant.

.evergreen/find-python3.sh
```diff
@@ -81,15 +81,30 @@ is_venv_capable() (
   local -r tmp="$(mktemp -d)"
   trap 'rm -rf "$tmp"' EXIT
 
-  "$bin" -m venv "$tmp" || return
+  if [[ "$OSTYPE" == cygwin ]]; then
+    local -r real_path="$(cygpath -aw "$tmp")" || return
+    "$bin" -m venv "$real_path" || return
+  else
+    "$bin" -m venv "$tmp" || return
```
ShaneHarvey, line 88: Can this be refactored to avoid duplicating `"$bin" -m venv "$tmp" || return`?

.evergreen/find-python3.sh
```diff
@@ -81,15 +81,30 @@ is_venv_capable() (
...
+  fi
+
+  # Sanity check: on some environments (such as Cygwin) creation of the virtual
+  # environment may succeed but place the environment in an unexpected location.
+  if [[ -n "$(find "$tmp" -maxdepth 0 -type d -empty 2>/dev/null)" ]]; then
```
ShaneHarvey, line 93: Can you show an example of this happening? Regardless can we remove this check because it's already handled by the if/elif/else below?

.evergreen/find-python3.sh
```diff
@@ -121,15 +136,30 @@ is_virtualenv_capable() (
...
+  else
+    "$bin" -m virtualenv -p "$bin" "$tmp" || return
+  fi
+
+  # Sanity check: on some environments (such as Cygwin) creation of the virtual
+  # environment may succeed but place the environment in an unexpected location.
+  if [[ -n "$(find "$tmp" -maxdepth 0 -type d -empty 2>/dev/null)" ]]; then
+    echo "$tmp is empty despite successful creation of virtual environment!"
+    return 1
+  fi
```
ShaneHarvey, line 151: Same comments as above.

.evergreen/csfle/activate-kmstlsvenv.sh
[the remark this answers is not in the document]
rcsanchez97: Oh, cool.  TIL

.evergreen/csfle/activate-kmstlsvenv.sh
```diff
@@ -5,24 +5,93 @@
...
+      # target` on macos-1012.
+      packages+=("greenlet<2.0")
+    fi
+
+    if [[ "$OSTYPE" == cygwin && "$HOSTTYPE" == x86_64 ]]; then
+      local -r windows_os_name="$(systeminfo.exe /FO LIST | perl -lne 'print $1 if m/^OS Name:\s+(.*)$/' || true)"
+
+      if [[ "$windows_os_name" =~ 2016 ]]; then
+        # Avoid `RuntimeError: Could not determine home directory.` on
+        # windows-64-2016. See BUILD-16233.
```
ShaneHarvey, line 49: Is this only windows-64-2016? What about windows-64-vsMulti-small (Microsoft Windows Server 2019 Datacenter)?

.evergreen/csfle/activate-kmstlsvenv.sh
```diff
@@ -5,24 +5,93 @@
...
+        python -m pip install -U "setuptools<65.0" || {
+          local -r ret="$?"
+          deactivate || return 1 # Deactivation should never fail!
+          return "$ret"
+        }
+      fi
+    fi
+
+    # Avoid `error: can't find Rust compiler`.
```
ShaneHarvey, line 58: What if instead of trying to pinpoint which platforms need cryptography<3.4 we just try to install the latest version and if that fails fallback to cryptography<3.4? Like this:
```
python -m pip install cryptography || python -m pip install 'cryptography<3.4' || ...

python -m pip install -U "${packages[@]}" || ...
```

This is simpler and should work on more platforms.

.evergreen/csfle/activate-kmstlsvenv.sh
ShaneHarvey: Update I reproduced the same issue there.

.evergreen/csfle/activate-kmstlsvenv.sh
eramongodb: It is indeed simpler, but I deliberately opted for the current approach in order to be _very_ explicit about conditions that require workarounds and as narrow as possible in the application of said workarounds.

This was motivated by the status quo where [generally-applied workarounds](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/4136a431f23461f1abe415e93d85fa55ec4c1be7/.evergreen/csfle/activate_venv.sh#L19) such as pinning cryptography to `~=3.4.8` or using `CRYPTOGRAPHY_DONT_BUILD_RUST=1` continued to demonstrate unexpected failures, and the conditions for said failures appeared to be inconsistent and opaque. It was unclear to me whenever I encountered such a failure whether it was already known, a new problem, or where the blame should be assigned (did _I_ break it, or did the environment change without my knowing?).

My hope was that being explicit in this manner would make it easier to maintain this script moving forward, with simplifications/removals of special-casing being applied in a controlled and targeted manner.

ShaneHarvey: I'd prefer the generic one to avoid needing to tweak and maintain these 30 extra lines which may or may not even cover all the hosts. I think a good compromise would be to use the generic approach but add an informative comment that specifically explains why the workaround exists like:
```
# Installing newer versions of cryptography requires rust when a wheel is not available.
# Fallback to an older version that does not require rust if the install fails. This is needed
# for at least the RHEL 6.2, powerpc64le, zSeries, and power8 hosts.
python -m pip install cryptography || python -m pip install 'cryptography<3.4' || ...
```

What do you think?

.evergreen/csfle/activate-kmstlsvenv.sh
ShaneHarvey: I'm not sure why windows-64-2019 works but windows-64-vsMulti-small doesn't. Either way the issue needs to be fixed on windows-64-vsMulti-small too because that's what we test on in pymongo.

.evergreen/find-python3.sh
ShaneHarvey: Yeah I would prefer removing it because it simplifies the script and we don't do anything special for an empty dir.

.evergreen/venv-utils.sh
ShaneHarvey: Oh, can you add the comment from the old script? It's much more informative.

ShaneHarvey (changes requested):

.evergreen/csfle/activate-kmstlsvenv.sh
eramongodb: I think that is an acceptable compromise. Would appeciate other reviewers' thoughts on this before committing to the refactor.

.evergreen/venv-utils.sh
eramongodb: Will do. 👍

.evergreen/csfle/activate-kmstlsvenv.sh
eramongodb: Just documenting what we discussed via other channels that the windows-64-vsMulti-small was [added to the test suite](https://spruce.mongodb.com/version/636ed44a3066155444043a3b/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) but [did not demonstrate failure](https://evergreen.mongodb.com/lobster/evergreen/task/drivers_tools_test_kmstlsvenv_windows_64_vsMulti_small_test_kmstlsvenv_patch_f5d1c049948718218b006c0732990e1c642cbb13_636ed44a3066155444043a3b_22_11_11_23_01_31/0/task#bookmarks=0%2C532&l=1&shareLine=515) that was observed when testing on a spawn host, and that this issue could be related to [BUILD-12392](https://jira.mongodb.org/browse/BUILD-12392).

.evergreen/csfle/activate-kmstlsvenv.sh
kevinAlbs: I have a slight preference for the compromise. That may require less changes to this script as distros undergo changes or more distros are added.

eramongodb: Done. Verified by [this patch](https://spruce.mongodb.com/version/6372b59c9ccd4e32ba9533e1).

ShaneHarvey (approved):

kevinAlbs (approved):

PR #247, opened 2022-11-22
DRIVERS-2465 Download latest Major release of `crypt_shared`

kevinAlbs:
# Summary

- Download latest Major release of `crypt_shared` if a version matching the server is unavailable.

After this change, downloading server versions < 6.0 is expected to try to download `crypt_shared` 6.0.

# Background & Motivation

The term Major comes from [MongoDB Software Lifecycle Schedules](https://www.mongodb.com/support-policy/lifecycles).

This is a request from product. Drivers only test the `crypt_shared` library with the same version of the server per DRIVERS-2355. MongoDB does not document support for using `crypt_shared` an older server version. Functionally, new versions of `crypt_shared` are expected to work with older server versions. Testing will validate that expectation and enable documenting support of new `crypt_shared` with old server versions.

Tested with [Go driver](https://spruce.mongodb.com/version/63a097815623430fa4eaf49e/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

### DIFF
--- a/.evergreen/download-mongodb.sh
+++ b/.evergreen/download-mongodb.sh
@@ -572,12 +572,16 @@ download_and_extract ()
       # Download 5.0 package to get the legacy mongo shell as a workaround until DRIVERS-2328 is addressed.
       echo "Legacy 'mongo' shell not detected."
       echo "Download legacy shell from 5.0 ... begin"
-      get_mongodb_download_url_for "$DISTRO" "5.0"
+      # Use a subshell to avoid overwriting MONGODB_DOWNLOAD_URL and MONGO_CRYPT_SHARED_DOWNLOAD_URL.
+      MONGODB50_DOWNLOAD_URL=$(
+         get_mongodb_download_url_for "$DISTRO" "5.0" > /dev/null
+         echo $MONGODB_DOWNLOAD_URL
+      )
 
       SAVED_DRIVERS_TOOLS=$DRIVERS_TOOLS
       mkdir $DRIVERS_TOOLS/legacy-shell-download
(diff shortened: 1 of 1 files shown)


DmitryLukyanov:
Consider [this](https://github.com/DmitryLukyanov/drivers-evergreen-tools/commit/c5108556141ea29fb23f8180dd1af2f8f0e28d33) simplification that doesn't require a new env

DmitryLukyanov (approved):
LGTM

.evergreen/download-mongodb.sh
```diff
@@ -525,9 +515,34 @@ get_mongodb_download_url_for ()
    fi
 
    if [ "$VERSION_INCLUDES_CRYPT_SHARED" = "YES" ]; then
+      # Use the same version as the server.
+      REQUESTED_CRYPT_SHARED_VERSION="$_VERSION"
```
DmitryLukyanov, line 519: maybe it makes sense to print what version we actually choose?

kevinAlbs: It is printed in [run-orchestration.sh](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/3b12064cc469fc18d8024d48353d29b89dd26472/.evergreen/run-orchestration.sh#L105-L117).

Also, I am unsure if there are callers of this script that rely on `get_mongodb_download_url_for` only printing the server URL. E.g. a caller may do this:
```sh
# Get the server download URL.
MONGODB_URL=$(get_mongodb_download_url_for $DISTRO $VERSION)
```

DmitryLukyanov (approved):
LGTM

PR #248, opened 2022-11-23
Migrate deprecated set-output command to use $GITHUB_OUTPUT

jmikola:
See: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

This eliminates deprecation notices in our workflow runs. Related to another PR I opened in https://github.com/cmb69/setup-php-sdk/pull/21, although that was for Windows.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -51,5 +51,5 @@ runs:
       shell: sh
     - id: "get-cluster-uri"
       name: "Expose Cluster URI"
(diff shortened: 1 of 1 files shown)


alcaeus (approved):
Thanks!

jmikola:
Testing in https://github.com/jmikola/mongo-php-driver/actions/runs/3531015098. I'll wait for the Ubuntu workflow runs to pass before merging this.

PR #254, opened 2022-12-15
Remove outdated bottle version cap

ShaneHarvey:
Testing for https://github.com/10gen/mongo-orchestration/pull/293 and removes the now outdated workaround added in https://github.com/mongodb-labs/drivers-evergreen-tools/pull/202


ShaneHarvey:
https://github.com/bottlepy/bottle/issues/1377 was fixed and released and MO no longer uses CherryPy so we can remove this version cap on bottle.

PR #255, opened 2022-12-16
Use find-python3.sh and venv-utils.sh to create mongo orchestration venv

eramongodb:
A recent upgrade to [use cheroot by default instead of CherryPy](https://github.com/10gen/mongo-orchestration/pull/293) to address Python 3.11 compatibility breaks the `run-orchestration.sh` script on platforms where the [manual Python binary lookup routine](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/10f1a128f09a80e4d4ac42579a0f17b81338ab81/.evergreen/start-orchestration.sh#L17) in `start-orchestration.sh` cannot find a suitably new version of Python, as observed in [this build](https://parsley.mongodb.com/evergreen/cxx_driver_debian10_release_50_compile_and_test_with_shared_libs_patch_7b4ceacb794c9e2953346d8dc9dae894315eaaf8_639ca1dd9ccd4e59daf572f8_22_12_16_16_50_37/0/task?bookmarks=0,295&selectedLine=217) (e.g. [selected a Python 2 binary](https://parsley.mongodb.com/evergreen/cxx_driver_debian10_release_50_compile_and_test_with_shared_libs_patch_7b4ceacb794c9e2953346d8dc9dae894315eaaf8_639ca1dd9ccd4e59daf572f8_22_12_16_16_50_37/0/task?bookmarks=0,295&selectedLine=83) on the `debian10-large` distro).

```
Traceback (most recent call last):
  File "/.../.evergreen/orchestration/venv/local/lib/python2.7/site-packages/mongo_orchestration/server.py", line 134, in run
    server=self.args.server)
  File "/.../.evergreen/orchestration/venv/bin/bottle.py", line 3172, in run
    server.run(app)
  File "/.../.evergreen/orchestration/venv/bin/bottle.py", line 2833, in run
    from cheroot import wsgi
  File "/.../.evergreen/orchestration/venv/local/lib/python2.7/site-packages/cheroot/wsgi.py", line 36, in <module>
    from . import server
  File "/.../.evergreen/orchestration/venv/local/lib/python2.7/site-packages/cheroot/server.py", line 87, in <module>
    from backports.functools_lru_cache import lru_cache
ImportError: No module named functools_lru_cache
```

This PR updates the Python binary lookup routine to use the new `find-python3.sh` script introduced in https://github.com/mongodb-labs/drivers-evergreen-tools/pull/236. This will encourage the `run-orchestration.sh` script to favor the newest version of Python available on any given distro.

PR #256, opened 2022-12-19
DRIVERS-2519: Utilize find-python3.sh in create-serverless.sh

jmikola:
https://jira.mongodb.org/browse/DRIVERS-2519

### DIFF
--- a/.evergreen/serverless/create-instance.sh
+++ b/.evergreen/serverless/create-instance.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -o errexit
 set +o xtrace # Disable xtrace to ensure credentials aren't leaked
--- a/.evergreen/serverless/create-instance.sh
+++ b/.evergreen/serverless/create-instance.sh
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
 
+if [ -z "$BASHPID" ]; then
+  echo "create-instance.sh must be run in a Bash shell!" 1>&2
+  exit 1
+fi
+
 set -o errexit
 set +o xtrace # Disable xtrace to ensure credentials aren't leaked
 
--- a/.evergreen/serverless/create-instance.sh
+++ b/.evergreen/serverless/create-instance.sh
@@ -46,6 +46,17 @@ if [ -z "$SERVERLESS_INSTANCE_NAME" ]; then
     SERVERLESS_INSTANCE_NAME="$RANDOM-DRIVERTEST"
 fi
 
+# Ensure that a Python binary is available for JSON decoding
+# shellcheck source=.evergreen/find-python3.sh
+. ../find-python3.sh || return
+
+PYTHON_BINARY="$(find_python3)"
+
+if [ -z "$PYTHON_BINARY" ]; then
+    echo "Failed to find Python3 binary"
+    exit 1
+fi
+
 echo "Creating new serverless instance \"$SERVERLESS_INSTANCE_NAME\"..."
 
 # See: https://www.mongodb.com/docs/atlas/reference/api/serverless/create-one-serverless-instance/
@@ -75,13 +86,6 @@ curl \
 EOF
 
 echo ""
-
-# Find the appropriate Python binary for JSON decoding
-if [ "Windows_NT" = "$OS" ]; then
-  PYTHON_BINARY=C:/python/Python38/python.exe
-else
-  PYTHON_BINARY=python3
-fi
 
 SECONDS=0
 DIR=$(dirname $0)
--- a/.evergreen/serverless/create-instance.sh
+++ b/.evergreen/serverless/create-instance.sh
@@ -51,16 +51,11 @@ if [ -z "$SERVERLESS_INSTANCE_NAME" ]; then
     SERVERLESS_INSTANCE_NAME="$RANDOM-DRIVERTEST"
 fi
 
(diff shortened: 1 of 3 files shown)


.evergreen/serverless/create-instance.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
jmikola, line 1: My understanding is that this is more portable. It's consistent with what @eramongodb did in other recent PRs.

.evergreen/serverless/create-instance.sh
```diff
@@ -46,6 +46,17 @@ if [ -z "$SERVERLESS_INSTANCE_NAME" ]; then
     SERVERLESS_INSTANCE_NAME="$RANDOM-DRIVERTEST"
 fi
 
+# Ensure that a Python binary is available for JSON decoding
+# shellcheck source=.evergreen/find-python3.sh
+. ../find-python3.sh || return
```
jmikola, line 51: This is related to my question in https://github.com/mongodb-labs/drivers-evergreen-tools/pull/255#discussion_r1052086238. I'm not sure why some other scripts (e.g. `start-orchestration.sh`) used `$DRIVER_TOOLS` instead of a relative path, but I think this is correct in this context.

Also, do I assume correctly that it's still preferable to use `find-python3.sh` even if we don't need to create a venv? This script only utilizes built-in Python modules (`sys` and `json`) so I don't think there's any need for a venv. I realize `find-python3.sh` requires the discovered Python environment to support venvs, but I assume that's not really a concern on Evergreen.

.evergreen/serverless/create-instance.sh
```diff
@@ -46,6 +46,17 @@ if [ -z "$SERVERLESS_INSTANCE_NAME" ]; then
...
+
+PYTHON_BINARY="$(find_python3)"
+
+if [ -z "$PYTHON_BINARY" ]; then
+    echo "Failed to find Python3 binary"
+    exit 1
+fi
```
jmikola, line 58: Previously, the script might have spawned a serverless instance via `curl` and _then_ failed due to a bad Python binary path. Moving this up just ensures we abort the script earlier before issuing any API requests.

eramongodb, line 58: I believe this can also be written as:
```bash
PYTHON_BINARY="$(find_python3)" || exit 1
```
Since `find_python3` already emits messages indicating when it is unable to find a binary.

eramongodb (approved):

.evergreen/serverless/create-instance.sh
eramongodb: That is the intention, yes. See https://unix.stackexchange.com/q/29608 for a deeper discussion on its merits and drawbacks.

.evergreen/serverless/create-instance.sh
jmikola: So long as I don't redirect stderr to `/dev/null`, that works we'll see output like the following:

> Could not find a python3 binary capable of creating a virtual environment!

I wasn't sure how this would work in practice, but https://stackoverflow.com/a/17336953/162228 explained the difference between `find_python3`'s exit code and the stdout output that we're capturing here. So `||` in your suggestion is actually branching on whether an empty string was returned, which is equivalent to the more verbose conditional I had here.

Thanks for the suggestion! I'll make that change.

.evergreen/serverless/create-instance.sh
```diff
@@ -46,6 +51,12 @@ if [ -z "$SERVERLESS_INSTANCE_NAME" ]; then
     SERVERLESS_INSTANCE_NAME="$RANDOM-DRIVERTEST"
 fi
 
+DIR=$(dirname $0)
```
jmikola, line 54: This line was just moved up from later in the file, as the current directory was previously used for invoking `get-instance.sh`.

I noted in the commit message that we could have used `${BASH_SOURCE[0]}` here but I don't think it makes any difference since this file is never sourced (related reading: https://stackoverflow.com/questions/35006457/choosing-between-0-and-bash-source).

.evergreen/serverless/delete-instance.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
```
jmikola, line 1: I used `shellcheck -s sh` to verify that neither of these scripts used bash syntax. Note that `get-instance.sh` is still invoked with `bash` from `create-instance.sh`. I'm open to removing that, but it seems harmless to leave as-is.

eramongodb: Personally I would prefer encouraging scripts to support and use Bash only, rather than the limited features of an inconsistent `sh` shell (Bourne? Bash? Dash?).

ShaneHarvey: I'd prefer leaving these as bash. It may not need bash syntax now but it might in the future and leaving it as bash avoids churn. Bash is also more consistent since it seems like we are moving towards requiring bash in all scripts.

.evergreen/serverless/create-instance.sh
jmikola: Noted. Per https://github.com/mongodb-labs/drivers-evergreen-tools/pull/256#discussion_r1052853294 I left the script's use of `$0` as-is since it is never sourced.

ShaneHarvey (approved):

PR #257, opened 2022-12-19
Add bash check and use BASH_SOURCE instead of DRIVERS_TOOLS

eramongodb:
This PR is a followup to https://github.com/mongodb-labs/drivers-evergreen-tools/pull/255 which assumed `DRIVERS_TOOLS` is defined when the script is invoked (as per `run-orchestration.sh`), which was not true (https://github.com/10gen/mongosync/pull/1180). This PR reverts to _not_ requiring `DRIVERS_TOOLS` in `start-orchestration.sh` by using `"$(dirname "${BASH_SOURCE[0]}")"` instead, taking advantage of the Bash shell requirement.

Furthermore, due to repeated instances of confusion regarding the Bash shell requirement and methods of invoking the script on Evergreen, this PR also introduces "Bash checks" to improve the quality of error messages that are encountered if a script is incorrectly invoked using i.e. `sh -c script.sh`, which does not respect the script shebangs ([sh is not bash!](https://mywiki.wooledge.org/BashGuide/CommandsAndArguments#Scripts)).

### DIFF
--- a/.evergreen/start-orchestration.sh
+++ b/.evergreen/start-orchestration.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-if [ -z "$BASH_SOURCE" ]; then
+if [ -z "$BASHPID" ]; then
   echo "start-orchestration.sh must be run in a Bash shell!" 1>&2
   exit 1
 fi
--- a/.evergreen/start-orchestration.sh
+++ b/.evergreen/start-orchestration.sh
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
 
+if [ -z "$BASH_SOURCE" ]; then
+  echo "start-orchestration.sh must be run in a Bash shell!" 1>&2
+  exit 1
+fi
+
 if [ "$#" -ne 1 ]; then
   echo "$0 requires one argument: <MONGO_ORCHESTRATION_HOME>"
   echo "For example: $0 /tmp/mongo-orchestration-home"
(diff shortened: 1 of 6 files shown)


.evergreen/start-orchestration.sh
```diff
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
 
+if [ -z "$BASHPID" ]; then
+  echo "start-orchestration.sh must be run in a Bash shell!" 1>&2
```
ShaneHarvey, line 4: Is this possible given the `#!/usr/bin/env bash` above?

eramongodb: Yes. Shebangs only apply if the script is executed without explicitly specifying the shell to use, i.e. `chmod +rx ./script.sh && ./script.sh`. Explicitly using `sh` as in `sh -c script.sh` does not respect the shebang (runs in `sh`, which is often _not_ the same as Bash).

ShaneHarvey (approved):

.evergreen/start-orchestration.sh
jmikola: Indeed. That tripped me up a few times while implementing downstream tickets for [DRIVERS-2497](https://jira.mongodb.org/browse/DRIVERS-2497) as our Evergreen config was still using `sh` to invoke the scripts that were recently changed to use these shebangs. Easily missed especially when there are a bunch of environment variables before the script path!

jmikola (approved):

PR #260, opened 2023-01-20
DRIVERS-2415 Implement OIDC SASL mechanism

blink1073:


.evergreen/auth_oidc/README.md
```diff
@@ -0,0 +1,41 @@
...
+and set the three AWS variables accordingly.
+
+. ./activate_venv.sh
+python oidc_write_orchestration.py
+python oidc_get_tokens.py
+
+This will create the tokens in `/tmp/tokens` and
+create the file `$DRIVERS_TOOLS/orchestration/configs/servers/auth-oidc.json`.
+
+You can then run mongo orchestration with `TOPOLOGY=server` and `ORCHESTRATION_FILE=auth-oidc.json`.
```
durran, line 35: Should this be `replica_set` instead of `server` here?

blink1073: Fixed

.evergreen/auth_oidc/setup_oidc.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Set up OIDC auth roles.
+ */
+(function() {
+"use strict";
+
+const admin = Mongo().getDB("admin");
+assert(admin.auth("bob", "pwd123"));
+assert.commandWorked(admin.runCommand({createRole: 'test1/readWrite', roles:[{role: 'readWrite', db: "test"}], privileges: []}));
+assert.commandWorked(admin.runCommand({createRole: 'test2/read', roles:[{role: 'read', db: "test"}], privileges: []}));
```
durran, line 10: Would it make sense to have a mongosh-friendly version of this? It uses Node's `assert` module so `assert.commandWorked` does not exist but we could get the equivalent in mongosh with:

```ts
  assert.deepEqual({ ok: 1 }, admin.runCommand({createRole: 'test1/readWrite', roles:[{role: 'readWrite', db: 'test'}], privileges: []}));
  assert.deepEqual({ ok: 1 }, admin.runCommand({createRole: 'test2/read', roles:[{role: 'read', db: 'test'}], privileges: []}));
```

blink1073: `assert.deepEqual` doesn't work with `mongo`, and it seems `mongosh` is not downloaded by `run-orchestration.sh`, unless I'm missing something?

durran: Correct, I was asking if it would be worth it to have a separate potential script, like `setup_oidc_mongosh.js` that could be optionally used. (Since `mongo` is EOL and mongo-orchestration needs to get a legacy binary). 

I just brought it up as a question - but nothing needs to happen here. I just did this in the Node driver for now because we'd prefer to use mongosh: https://github.com/mongodb/node-mongodb-native/pull/3560/files#diff-58d5ff1e7dd36ee2f194daa1bb078a421899232c908c8e9bbe50c5401f0c2d02

blink1073: Perhaps instead we should update `run-orchestration.sh` to install `mongosh` as well, and only use `mongosh` in this instance?  That would make it easier to update other usages of `mongo` in this repo.

durran: That sounds like a good idea.

blink1073: Actually, that's a lot more complicated than I thought, given the need to support multiple platforms.  I'll defer to a dedicated effort and add the second script.

durran:
@blink1073 Should this be rebased against master to pick up https://github.com/mongodb-labs/drivers-evergreen-tools/commit/6ac2fdd5eadc0a9deb960c661c3f9ecca1b0cca9 ?

blink1073:
I got the following, it seems like we can rely on `mongosh` to raise an [error](https://www.mongodb.com/docs/mongodb-shell/run-commands/#command-exceptions) instead of manually checking the response?

```
AssertionError: Expected values to be loosely deep-equal:

{
  ok: 1
}

should loosely deep-equal

{
  '$clusterTime': {
    clusterTime: t {
      high: 1676315481,
      low: 3,
      unsigned: true
    },
    signature: {
      hash: e {
        buffer: Buffer(20) [Uint8Array] [
          87,
          53,
          60,
          95,
          104,
          172,
          139,
          103,
          200,
          159,
          251,
          50,
          214,
          254,
          184,
          62,
          20,
          219,
          63,
          252
        ],
        position: 20,
```

durran:
> I got the following, it seems like we can rely on `mongosh` to raise an [error](https://www.mongodb.com/docs/mongodb-shell/run-commands/#command-exceptions) instead of manually checking the response?
> 
> ```
> AssertionError: Expected values to be loosely deep-equal:
> 
> {
>   ok: 1
> }
> 
> should loosely deep-equal
> 
> {
>   '$clusterTime': {
>     clusterTime: t {
>       high: 1676315481,
>       low: 3,
>       unsigned: true
>     },
>     signature: {
>       hash: e {
>         buffer: Buffer(20) [Uint8Array] [
> ...
> ```

Yeah I think checking the response isn't needed here.

.evergreen/find-python3.sh
```diff
@@ -150,7 +150,7 @@ is_virtualenv_capable() (
     # shellcheck source=/dev/null
     . "$tmp/bin/activate"
   elif [[ -f "$tmp/Scripts/activate" ]]; then
-    dos2unix "$tmp/Scripts/activate" || return
```
blink1073, line 153: @eramongodb this command was erroring out locally on @DmitryLukyanov's Windows machine.  I don't think it is strictly necessary, do you disagree?

eramongodb: What was the nature of the error (missing `dos2unix`? an error during conversion?)?

This command is inherited from [older venv creation/activation scripts](https://github.com/mongodb-labs/drivers-evergreen-tools/blame/0638b75db86aae603fa882271a251e86309fe050/.evergreen/csfle/activate_venv.sh#L21) to address a [still-unresolved venv issue in Cygwin environments](https://bugs.python.org/issue32451) (context documented in [venv-utils.sh](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/9eff0be763b12b4417ca035dd18eb5170c9e4336/.evergreen/venv-utils.sh#L72)) where the activation script fails during execution/sourcing due to:
```
line 3: $'\r': command not found
```

This issue is still relevant on [many Windows Evergreen distros](https://spruce.mongodb.com/version/63f6668357e85a38474bdf1a/tasks) despite the introduction of "prefer newer Python versions" and virtualenv fallback behavior (note: despite successful virtual environment creation, logs reveal scripts are forced to select old Python binaries + fallback to virtualenv despite availability of newer Python binaries + venv, which also affects availability/selection of packages). Unfortunately, although this issue was [resolved by virtualenv in 20.0.33](https://github.com/pypa/virtualenv/issues/1818), it remains unresolved for venv, thus the manual `dos2unix` workaround was still deliberately included in the new creation/activation scripts to enable greater accessibility and use of venv across Evergreen distros.

We could consider demoting the error handling from `|| return` to `|| true` if a benign conversion error is preventing proper script execution in local environments, but given these scripts are designed to be executed in Evergreen environments, I suggest simply commenting it out locally as a quick workaround if the issue is also exclusive to the local machine.

blink1073: I believe the command was present but failing in this instance.  In the case of OIDC, there is cause to run these scripts quite often locally while writing the tests.  I'd be in favor of using `|| true`.

eramongodb: Created https://github.com/mongodb-labs/drivers-evergreen-tools/pull/270 in response to this issue.

.evergreen/venv-utils.sh
```diff
@@ -38,20 +38,20 @@ fi
 # they may be deleted as part of virtual environment creation.
 venvcreate() {
   local -r bin="${1:?'venvcreate requires a Python binary to use for the virtual environment'}"
-  local -r path="${2:?'venvcreate requires a path to the virtual environment to create'}"
+  local -r venv_path="${2:?'venvcreate requires a path to the virtual environment to create'}"
```
eramongodb, line 41: Curious, what motivated the renaming of this variable? Was it interfering/conflicting with another `$path` variable?

durran (changes requested):

.evergreen/venv-utils.sh
blink1073: Yes, that is correct

durran (approved):

PR #262, opened 2023-01-24
Increase 60 server version.

DmitryLukyanov:
EG: https://evergreen.mongodb.com/version/63d05fd53627e01c122d4887


ShaneHarvey (approved):

ShaneHarvey:
Should we update rapid to 6.2.0-rc6?

DmitryLukyanov:
> Should we update rapid to 6.2.0-rc6?

I would love to (we need this change too). But not sure whether we can do it now. As I know, rapid releases land every quarter. So we need to do it soon, but don't know whether we can do it right now

ShaneHarvey:
Let's do it.

ShaneHarvey (approved):

JamesKovacs:
Rapid releases are intended to test our drivers against the current rapid release on Atlas, which is 6.1.1 (based on https://www.mongodb.com/download-center/enterprise/releases/development). Should we be testing our drivers against a rapid release RC? What benefit does that provide as we are already testing on `latest` already?

alcaeus:
I agree with @JamesKovacs here. `rapid` should test what Atlas is giving us. I would suggest adding `unstable` or `rc` that would allow testing with 6.2-rc now while keeping 6.1 tests through `rapid`. Once 6.2 is released we can then switch the `unstable` branch to 6.3-rc and test that version.
Note that I'm not suggesting drivers actually test against unstable unless they need to (e.g. for queryable encryption features in this case), but this way we would make it available while retaining the current versions.

DmitryLukyanov:
> I still think we should bump to rc6 here. The rationale is that we really need to test the next rapid release otherwise 6.2.0 will be released without ever being tested by drivers. We already know 6.1 works so testing 6.2.0-rc is more important.



> I still think we should bump to rc6 here. The rationale is that we really need to test the next rapid release otherwise 6.2.0 will be released without ever being tested by drivers. We already know 6.1 works so testing 6.2.0-rc is more important.



> I still think we should bump to rc6 here. The rationale is that we really need to test the next rapid release otherwise 6.2.0 will be released without ever being tested by drivers. We already know 6.1 works so testing 6.2.0-rc is more important.

It probably makes sense to do it as separate work. Discussed offline. In this PR, we will stick with 6.1.1 rapid version.

DmitryLukyanov:
> I still think we should bump to rc6 here. The rationale is that we really need to test the next rapid release otherwise 6.2.0 will be released without ever being tested by drivers. We already know 6.1 works so testing 6.2.0-rc is more important.

It probably makes sense to do it as separate work. Discussed offline. In this PR, we will stick with 6.1.1 rapid version.

ShaneHarvey (approved):

PR #266, opened 2023-02-14
PYTHON-3579 Add flag to prevent downloading crypt_shared when not needed

juliusgeo:


juliusgeo:
Because we only want to download crypt_shared in certain circumstances, this predicates the downloading of crypt_shared on whether the env var TEST_CRYPT_SHARED is defined.

.evergreen/run-orchestration.sh
```diff
@@ -101,21 +101,21 @@ URI=$(python -c 'import sys, json; j=json.load(open("tmp.json")); print(j["mongo
...
-else
-  echo "Downloading crypt_shared package from $MONGO_CRYPT_SHARED_DOWNLOAD_URL"
-  download_and_extract_crypt_shared "$MONGO_CRYPT_SHARED_DOWNLOAD_URL" "$EXTRACT" CRYPT_SHARED_LIB_PATH
-  echo "CRYPT_SHARED_LIB_PATH:" $CRYPT_SHARED_LIB_PATH
-  if [ -z $CRYPT_SHARED_LIB_PATH ]; then
-    echo "CRYPT_SHARED_LIB_PATH must be assigned, but wasn't" 1>&2 # write to stderr"
-    exit 1
-  fi
-cat <<EOT >> mo-expansion.yml
+if [ -n "$TEST_CRYPT_SHARED" ]; then
```
blink1073, line 104: Only the Python driver uses this variable.  We should add a new flag whose existence indicates not to download crypt shared.

blink1073 (changes requested):

ShaneHarvey:
Instead of adding a new flag could we remove the broken Amazon 2018 download link?

.evergreen/run-orchestration.sh
```diff
@@ -101,21 +101,21 @@ URI=$(python -c 'import sys, json; j=json.load(open("tmp.json")); print(j["mongo
...
-else
-  echo "Downloading crypt_shared package from $MONGO_CRYPT_SHARED_DOWNLOAD_URL"
-  download_and_extract_crypt_shared "$MONGO_CRYPT_SHARED_DOWNLOAD_URL" "$EXTRACT" CRYPT_SHARED_LIB_PATH
-  echo "CRYPT_SHARED_LIB_PATH:" $CRYPT_SHARED_LIB_PATH
-  if [ -z $CRYPT_SHARED_LIB_PATH ]; then
-    echo "CRYPT_SHARED_LIB_PATH must be assigned, but wasn't" 1>&2 # write to stderr"
-    exit 1
-  fi
-cat <<EOT >> mo-expansion.yml
+if [ -z ${SKIP_CRYPT_SHARED+''} ]; then
```
ShaneHarvey, line 104: What is the purpose of `+''`?

ShaneHarvey (changes requested):

.evergreen/run-orchestration.sh
juliusgeo: It is normally used when you want a default value for a variable expansion: ${SKIP_CRYPT_SHARED+'default value if variable is not defined'}. However, in this case we do not need it, so I removed it.

ShaneHarvey: Oh in that case you should use this style which is what we use elsewhere:
```
if [ -z "${SKIP_CRYPT_SHARED:-}" ]; then
```

ShaneHarvey: Please also add a comment that says something like:
```
# Define SKIP_CRYPT_SHARED=1 to skip downloading crypt_shared, this is useful 
# for platforms that have a server build but don't ship a corresponding crypt_shared
# release, like Amazon 2018.
```

juliusgeo: Done.

ShaneHarvey (approved):

PR #276, opened 2023-03-14
Update EG Matrix MongoDB 3.6+

blink1073:
Follow up to https://github.com/10gen/mongo-orchestration/pull/301

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1002,7 +1002,7 @@ buildvariants:
      - name: "test-latest-standalone"
      - name: "test-latest-load_balancer"
 
-# Storage Engine Tests on Amazon Linux (Enterprise)
+# Storage Engine Tests on Ubuntu 18
 - matrix_name: "tests-storage-engines"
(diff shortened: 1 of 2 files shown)


ShaneHarvey:
The evergreen PR task says:
> evergreen — no tasks/variants were configured

Could you update the tasks in the evg project settings page?

.evergreen/config.yml
```diff
@@ -968,24 +807,12 @@ buildvariants:
      - name: "test-latest-sharded_cluster"
      - name: "test-latest-standalone"
      - name: "test-latest-load_balancer"
+     - name: "test-6.0-replica_set"
+     - name: "test-6.0-sharded_cluster"
+     - name: "test-6.0-standalone"
```
ShaneHarvey, line 812: What about 4.0 - 5.0? Should we add those too?

ShaneHarvey (changes requested):

.evergreen/config.yml
blink1073: Done

ShaneHarvey (approved):

PR #275, opened 2023-03-14
DRIVERS-2415 Wait for replicaset primary to be available

blink1073:

PR #277, opened 2023-03-15
DRIVERS-2415 Better wait for replicaset

blink1073:
Follow up to #275 to avoid spamming the server.

### DIFF
--- a/.evergreen/auth_oidc/setup_oidc.js
+++ b/.evergreen/auth_oidc/setup_oidc.js
@@ -10,6 +10,16 @@ admin.runCommand({createRole: 'test1/readWrite', roles:[{role: 'readWrite', db: 
 admin.runCommand({createRole: 'test2/read', roles:[{role: 'read', db: 'test'}], privileges: []});
 
 // Wait for the node to be primary.
-while (admin.runCommand('replSetGetStatus').myState != 1) {}
+var shouldQuit = false;
+function waitForReady() {
+    if (admin.runCommand('replSetGetStatus').myState != 1) {
+        setTimeout(waitForReady, 1000);
+    } else {
+        console.log("ReplicaSet is Configured!");
+        shouldQuit = true;
+    }
+}
+waitForReady();
+while (!shouldQuit) {}
 
 }());


.evergreen/auth_oidc/setup_oidc.js
```diff
@@ -10,6 +10,16 @@ admin.runCommand({createRole: 'test1/readWrite', roles:[{role: 'readWrite', db:
...
+function waitForReady() {
+    if (admin.runCommand('replSetGetStatus').myState != 1) {
+        setTimeout(waitForReady, 1000);
+    } else {
+        console.log("ReplicaSet is Configured!");
+        shouldQuit = true;
+    }
+}
+waitForReady();
+while (!shouldQuit) {}
```
ShaneHarvey, line 23: This is still a tight CPU loop.

ShaneHarvey (changes requested):

.evergreen/auth_oidc/setup_oidc.js
ShaneHarvey: I think you can use:
```
while (admin.runCommand('replSetGetStatus').myState != 1) {
    sleep(50);
}
```

blink1073: Ah, `sleep()` is much better.  If only that were built into JS itself...

.evergreen/auth_oidc/setup_oidc.js
```diff
@@ -10,6 +10,10 @@ admin.runCommand({createRole: 'test1/readWrite', roles:[{role: 'readWrite', db:
 admin.runCommand({createRole: 'test2/read', roles:[{role: 'read', db: 'test'}], privileges: []});
 
 // Wait for the node to be primary.
-while (admin.runCommand('replSetGetStatus').myState != 1) {}
+while (admin.runCommand('replSetGetStatus').myState != 1) {
+    sleep(50);
+}
+
+exit();
```
ShaneHarvey, line 17: Is the `exit()` intentional?

blink1073: Yes, otherwise it was hanging for several seconds before exiting.

PR #278, opened 2023-03-20
DRIVERS-2415 Ensure 27017 is the Primary in the OIDC Replica Set

blink1073:
cc @DmitryLukyanov 

Avoids error when the replica set has not yet been configured.

### DIFF
--- a/.evergreen/auth_oidc/oidc_write_orchestration.py
+++ b/.evergreen/auth_oidc/oidc_write_orchestration.py
@@ -51,10 +51,7 @@ def main():
                     "oidcIdentityProviders": providers,
                     "featureFlagOIDC": True
                 }
-            },
-            "rsParams": {
-                "priority": 99
-            },
+            }
         }]
     }
 
@@ -86,7 +83,7 @@ def main():
             }
         },
         "rsParams": {
-            "priority": 1
+            "priority": 0
         }
     })
 
--- a/.evergreen/auth_oidc/oidc_write_orchestration.py
+++ b/.evergreen/auth_oidc/oidc_write_orchestration.py
@@ -39,8 +39,8 @@ def main():
         "login": "bob",
         "name": "mongod",
         "password": "pwd123",
-        "members": [
-            {"procParams": {
+        "members": [{
+            "procParams": {
                 "ipv6": "NO_IPV6" not in os.environ,
                 "bind_ip": "0.0.0.0,::1",
                 "logappend": True,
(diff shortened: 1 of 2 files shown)


ShaneHarvey:
Why is the replica set not configured at this point? Are we not using MO to start the set?

blink1073:
We are using MO to start the cluster, but `mongosh` was still sometimes giving `MongoServerError: not primary`.

ShaneHarvey:
That sounds like a bug in mongosh since 1) MO already waits for the primary to be available and 2) mongosh should be waiting in server selection for the primary to be online. Could you follow up with the node team?

blink1073:
It shows the primary as being in state 2 in the MO log.

ShaneHarvey:
State 2 is the secondary state:

https://www.mongodb.com/docs/manual/reference/replica-states/#replica-set-member-states

Does this script (incorrectly) expect the primary to be running on a specific port? It should work as long as any one of the nodes is the primary.

blink1073:
Good catch, the prose tests rely on 27017 being the primary.  I've added a priority setting.

ShaneHarvey:
Why do they rely on that being the primary?

blink1073:
Because one server is configured with one OIDC provider and the second is configured with two OIDC providers, which have differing behavior.

DmitryLukyanov:
Still [failing](https://evergreen.mongodb.com/task_log_raw/dot_net_driver_oidc_auth_test__version~latest_os~ubuntu_2004_topology~replicaset_auth~auth_ssl~ssl_oidc_auth_tests_patch_8f8f7e0eccb8da16f76aed463d27c54633ed836b_6418d9f6d6d80a0b0e5d6724_23_03_20_22_11_18/3?type=T#L1121) :(  Launched 8 mins ago

DmitryLukyanov:
Sounds like it's fixed on my side now

blink1073:
> Is there a use case for having different OIDC provider config on members in the same cluster? Should we start two different clusters instead?

As in, launch MO, run a subset of tests, then launch MO again?

ShaneHarvey:
Yes, we could run multiple clusters in sequence, run them in parallel, split the testing matrix in two, or keep the current setup. What use case are we testing by using a different OIDC config on a secondary?

blink1073:
When there is a single principal configured, the `username` is optional.  It is not optional when there are multiple principals.  The reason to prefer a single host is that we planned to move OIDC to a dedicated host, and we'd have to manage two hosts.

ShaneHarvey (approved):

ShaneHarvey:
I'm not going to hold up this work but I do think it would be good to move away from testing with different configs on replica set members since that's unexpected and will cause issues if/when there are elections.

PR #281, opened 2023-03-23
Add DPkg::Lock::Timeout to apt-get install commands

eramongodb:
This PR is motivated by the occasional observation of the following error (by multiple Drivers teams) when running Azure KMS tests:

```
E: Could not get lock /var/lib/dpkg/lock-frontend - open (11: Resource temporarily unavailable)
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?
```

This PR proposes adding the [undocumented(?)](https://manpages.ubuntu.com/manpages/jammy/man5/apt.conf.5.html) `DPkg::Lock::Timeout` option when invoking `apt-get install` to use its built-in support for [waiting on the dpkg lock](https://salsa.debian.org/apt-team/apt/-/merge_requests/109/diffs?commit_id=93e1565796b61eb44bec39f50e09a34cbe090178) instead of triggering an immediate error. This is probably preferable to [bypassing the lock](https://manpages.ubuntu.com/manpages/jammy/en/man1/dpkg.1.html) via `DPKG_FRONTEND_LOCKED`.

In `apt-get` versions [prior to 1.9.11](https://salsa.debian.org/apt-team/apt/-/commit/de9762d5443b8a957fae781bbdc4a740cc6c4e6f), the option will simply be ignored by the command. This applies to the ubuntu1404, ubuntu1604, ubuntu1804, debian81, and debian91 distros.

Note this option only applies to the `apt-get install` program. There is the possibility that `apt-get update` may also fail when attempting to acquire it's own apt lock (separate from the dpkg lock). This can be addressed by setting the `Debug::NoLocking=true` option if deemed necessary.

Also note that the timeout is currently set to `-1` to wait indefinitely for the dpkg lock to be released. If we do not want to wait indefinitely, we can set an explicit value (in seconds) instead.

### DIFF
--- a/.evergreen/auth_oidc/Dockerfile
+++ b/.evergreen/auth_oidc/Dockerfile
@@ -1,6 +1,6 @@
 FROM ubuntu:20.04
 
-RUN apt-get update -qq && apt-get install -qqy --no-install-recommends \
+RUN apt-get -qq update && apt-get -qqy -o DPkg::Lock::Timeout=-1 install --no-install-recommends \
   git \
   ca-certificates \
   curl \
--- a/.evergreen/csfle/azurekms/install-az.sh
+++ b/.evergreen/csfle/azurekms/install-az.sh
@@ -8,7 +8,7 @@ set -o nounset
 echo "Install az ... begin"
 # Once a sufficient version of `az` is installed on all tested distros, remove the install-az.sh script. BUILD-16836 requests installation of the latest `az` on supported distros.
 sudo apt-get update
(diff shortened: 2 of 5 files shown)


blink1073 (approved):
Thanks!

kevinAlbs (approved):
LGTM. Thank you for the improvement.

> Also note that the timeout is currently set to `-1` to wait indefinitely for the dpkg lock to be released. If we do not want to wait indefinitely, we can set an explicit value (in seconds) instead.

`-1` seems OK. I expect the Evergreen task timeout will put an upper bound on the wait time.

PR #286, opened 2023-03-29
fix: ADL configuration requires hostRegion setting

nbbeeken:
ADL configuration requires `hostRegion` setting


W-A-James (approved):
Tested on spawn host, fixes the issue.

LGTM

PR #295, opened 2023-04-13
DRIVERS-2605 Update aws_e2e_ec2.js for build team AWS EC2 changes

markbenvenuto:
Evergreen EC2 machines have a role assigned to them by default. As a result, the expected user name is different name for the assigned role.

CC: @blink1073 @kevinAlbs


blink1073 (approved):
Thank you!

PR #305, opened 2023-05-08
DRIVERS-2625 Use curl --retry-all-errors when it's supported

ShaneHarvey:
https://jira.mongodb.org/browse/DRIVERS-2625


blink1073 (approved):
Thanks!

PR #306, opened 2023-05-09
DRIVERS-2328 Use Python and Mongo Shell for AWS Testing

blink1073:
The script can be called as follows:

```bash
cd .evergreen/auth_aws
. ./activate-authawsvenv.sh
python aws_tester.py ec2  # ecs, assume-role, regular, web-identity
```

Tested with the Python driver with all 5 variants here: https://spruce.mongodb.com/task/mongo_python_driver_aws_auth_test__platform~ubuntu_18.04_python_version~3.7_aws_auth_test_6.0_patch_873032660bf22d09cbf0013f6b077196d0b95f40_645aaaa63e8e86cbdcbd8933_23_05_09_20_18_56/logs?execution=1&sortBy=STATUS&sortDir=ASC

Linked PR: https://github.com/mongodb/mongo-python-driver/pull/1220


blink1073:
cc @dariakp

durran:
@blink1073 what are the proposed driver changes supposed to look like? I did a commit to run an Evergreen patch on the Node driver against your fork/branch to see what this looked like and I get some AWS tests passing, but some failing as well.

My driver changes: https://github.com/mongodb/node-mongodb-native/commit/57c3f6573d03ea64494114ad8c7e9d67a9c81a52

My patch: https://spruce.mongodb.com/version/645bf0c856234354a5daaec9/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

blink1073:
Here's my patch:

https://github.com/mongodb/mongo-python-driver/compare/mongodb:mongo-python-driver:master...blink1073:mongo-python-driver:PYTHON-3692

In the ones that are passing in your patch, I see the expected "Creating user" output in the log.

Perhaps set each command to use "bash" and add a "set -ex" so we can debug?

ShaneHarvey:
@markbenvenuto would you like to merge any of this work back into the server test suite so that we can keep parity?

alcaeus:
I tested this using the .NET driver (I used that for all my testing as PHP does not implement these tests itself), and the only test that failed was the ecs test. However, the .NET driver has some special logic there to move the test file around, so I'd attribute the failure to my setup rather than assume it was the fault of these changes. I'll wait for @durran to run a new patch with debug output.

durran:
I cleaned up the patch to just test latest and added debug, only passing 3 of the 8. https://spruce.mongodb.com/version/645d33893627e020f687a8a5/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

I haven't dug into the cause as of yet, but my changes look consistent with the Python changes. This is the latest diff: https://github.com/mongodb/node-mongodb-native/commit/9584639d46dec4d44f4509013ec02fa07690142a

blink1073:
You need to add `shell: bash` whenever you are calling `. ./activate-authawsvenv.sh`.

durran:
@blink1073 Perfect - just the ECS failure now: https://spruce.mongodb.com/task/mongo_node_driver_next_ubuntu1804_test_mongodb_aws_aws_latest_auth_test_run_aws_ECS_auth_test_patch_64dc5771106d04de700796f0caea58d96833aa7a_645d4426d1fe072dd5475735_23_05_11_19_38_46/logs?execution=0

blink1073:
@durran you need to use `export PROJECT_DIRECTORY="${ECS_SRC_DIR}";`

durran (approved):
From a functional standpoint these work on the Node branch now. I'll let someone else review the Python code itself.

alcaeus (commented):
Changes LGTM as well. I ran into failures when testing this with the C# driver but I assume it's connected to the setup (which I'm not entirely familiar with) rather than the tests itself.

What I didn't test was whether this will have a breaking impact on drivers, i.e. do drivers have to modify their evergreen config as soon as we merge this, or is this something they can opt into?

PR #310, opened 2023-05-25
DRIVERS-2328 Fix Windows Support for AWS Scripts

blink1073:
Tested with https://github.com/mongodb/mongo-python-driver/pull/1220, and with a patch build that uses this branch without changing any EG config.


blink1073:
Tested withhttps://github.com/mongodb/mongo-csharp-driver/pull/1097 as well

ShaneHarvey (approved):

PR #314, opened 2023-06-02
Add dedicated perf version tag for server 6.x

dariakp:
Adds a dedicated tag `v6.0-perf` pinned to 6.3.1 (same as current rapid).

Once 7.0 is GA, we'll want to add a similar 7.0 perf version that drivers can use to test against in addition to this version.

Node driver [test run](https://spruce.mongodb.com/task/mongo_node_driver_next_performance_tests_run_spec_benchmark_tests_patch_ed5e27b77dd3393ba845fcead5f0baefc10add56_647a44e8562343aa2beb0913_23_06_02_19_37_19/logs?execution=4)

### DIFF
--- a/.evergreen/download-mongodb.sh
+++ b/.evergreen/download-mongodb.sh
@@ -3,6 +3,7 @@
 #For future use the feed to get full list of distros : http://downloads.mongodb.org/full.json
 
 set -o errexit  # Exit the script with error if any of the commands fail
+set -o xtrace
 
 get_distro ()
 {
--- a/.evergreen/download-mongodb.sh
+++ b/.evergreen/download-mongodb.sh
@@ -3,7 +3,6 @@
 #For future use the feed to get full list of distros : http://downloads.mongodb.org/full.json
 
 set -o errexit  # Exit the script with error if any of the commands fail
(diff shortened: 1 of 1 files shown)


.evergreen/download-mongodb.sh
```diff
@@ -58,6 +58,8 @@ get_mongodb_download_url_for ()
    VERSION_70="7.0.0-rc0"
    VERSION_60_LATEST="v6.0-latest"
    VERSION_60="6.0.6"
+   # The perf version must always remain pinned to the same patch
+   VERSION_60_PERF="6.3.1"
```
ShaneHarvey, line 62: Should this be 6.0.6 since it's VERSION_60_PERF?

dariakp: Yeah that's a good point, I was autopiloting to match what node is currently testing against, which is rapid, but I think it makes the most sense to use the "official" current version instead, and that's what we'll likely be adding for 7.0 testing going forward, too.

blink1073 (approved):
Thanks!

ShaneHarvey (approved):

PR #315, opened 2023-06-09
Clean up download links

blink1073:
- Use stable release branches to pick up patch releases automatically.
- Prune unsupported server versions and platforms.

### DIFF
--- a/.evergreen/download-mongodb.sh
+++ b/.evergreen/download-mongodb.sh
@@ -43,7 +43,7 @@ get_distro ()
    echo $DISTRO
 }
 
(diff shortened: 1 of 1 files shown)


matthewdale (approved):
Looks good 👍

ShaneHarvey:
Unfortunately we should revert this change. The problem with using these links is that they point to the lasted _build_ of server on that branch, not the latest stable release. For example v3.6 on ubuntu1804 points to v3.6.23-8-gc2609ed3ed, not 3.6.23:
```
 [2023/06/09 15:05:32.361] http://downloads.10gen.com/linux/mongodb-linux-x86_64-enterprise-ubuntu1804-v3.6-latest.tgz
 [2023/06/09 15:05:37.463] db version v3.6.23-8-gc2609ed3ed
```

PR #316, opened 2023-06-09
PYTHON-3721 Make retry more robust

NoahStapp:
curl's --retry-all-errors is not available on versions older than 7.71.0, and the older --retry option does not retry non-transient server errors such as 4xx and 5xx. By using the --fail option and manually checking the status code of the completed curl command, this fix ensures that the `curl_retry` function retries all failures up to three times.

### DIFF
--- a/.evergreen/download-mongodb.sh
+++ b/.evergreen/download-mongodb.sh
@@ -635,14 +635,11 @@ get_mongodb_download_url_for ()
    echo $MONGODB_DOWNLOAD_URL
 }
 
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
LGTM, thanks!  (We're still waiting on a passing patch build in PyMongo with these changes before merging).

NoahStapp:
Patch build succeeded: https://spruce.mongodb.com/version/64834f4b850e6169f9556bbf/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

PR #331, opened 2023-07-26
DRIVERS-2384 Script Clean Up

blink1073:
- Remove portions of `run-deployed-lambda-aws-tests.sh` that have moved to setup/teardown scripts
- Enforce input variables
- Handle function failures and exit accordingly
- Actually show the script logs

Tested with https://github.com/mongodb/mongo-python-driver/pull/1310


ShaneHarvey (approved):
LGTM! Note Durran is out until next week.

blink1073:
I'll make a PR for Go as well (I'm the lead on that team).  I had already made https://github.com/mongodb/mongo-go-driver/pull/1329 to start using the individual scripts.

blink1073:
Okay, this is now passing in https://github.com/mongodb/mongo-go-driver/pull/1343/files as well.  I didn't catch that there were two versions of `run-deployed-lambda-aws-tests.sh`.  I think we should remove the old one in favor of the one in `aws_lamba` now that no one is using it.

ShaneHarvey (commented):
>  I think we should remove the old one in favor of the one in aws_lamba now that no one is using it.

SGTM should we do that in this PR?

blink1073:
I think we should remove it in this PR, WDYT @durran?

durran:
> I think we should remove it in this PR, WDYT @durran?

Yes it was intentional to remove it and keep the one in aws_lambda so go for it.

PR #332, opened 2023-07-28
Add option to get debug binaries in get_mongodb_download_url_for

jddubois:
This change enables mongosync integration and e2e tests to retrieve debug symbol binaries for testing- these binaries are uploaded to evergreen for easy access. This will improve the debugging experience for the core server team when mongosync exposes server bugs.

Note that not all of these debug URLs are valid, but most are. This change should not have an impact on existing usage of `get_mongodb_download_url_for` but debug symbols don't seem to exist for some versions and distros.


matthewdale (approved):
Looks good 👍

PR #336, opened 2023-08-14
DRIVERS-2415 Add SCRAM-SHA-1 to OIDC Server

blink1073:


blink1073:
Tested here: https://spruce.mongodb.com/task/mongo_python_driver_oidc_auth_test__platform~rhel8_python_version~3.9_oidc_auth_test_latest_patch_c1d33831e8137eb0715d7531baf5b86758bb9187_64da5b5fe3c331e9dc28294c_23_08_14_16_50_50?execution=0&sortBy=STATUS&sortDir=ASC

ShaneHarvey (changes requested):
I don't think this is needed. Drivers default to using SCRAM-SHA-256 when the user supports it.

PR #339, opened 2023-08-21
DRIVERS-2411 delete Azure Virtual Network

kevinAlbs:
# Summary

- Add `az network vnet delete` to delete created Virtual Network.
- Add `--vnet-name` option to the `azure vm create` command.

Tested with C driver: https://spruce.mongodb.com/version/64e3a22ad1fe07741da84c88/tasks

# Background & Motivation
`azure vm create` creates a Virtual Network along with the Virtual Machine. The Virtual Network does not appear to be deleted with `azure vm delete`. After running `delete-vm.sh`, a Virtual Network was observed in the portal:

Leftover Virtual Network


kevinAlbs:
Disregard. The Virtual Network created appears to be scoped to the Resource Group. This PR may result in creating a Virtual Network per VM, rather than per Resource Group.

PR #343, opened 2023-08-29
Make EVG outputs quieter

blink1073:
Make the EVG outputs quieter:

- Use `$(find_python3 2>/dev/null)` in more places
- Use `pip install -q` for quieter pip installs
- Avoid error when trying to shut down an HAProxy that was not running
- Remove legacy code in venv scripts supporting Windows 2016

### DIFF
--- a/.evergreen/auth_aws/activate-authawsvenv.sh
+++ b/.evergreen/auth_aws/activate-authawsvenv.sh
@@ -33,11 +33,12 @@ activate_authawsvenv() {
     # shellcheck source=.evergreen/find-python3.sh
     . ../find-python3.sh || return
 
-    echo "Creating virtual environment 'authawsvenv'..."
     echo "Finding Python3 binary..."
-    PYTHON="$(find_python3 2>/dev/null)"
+    PYTHON="$(find_python3 2>/dev/null)" || return
     echo "Finding Python3 binary... done."
-    venvcreate $PYTHON authawsvenv || return
+
+    echo "Creating virtual environment 'authawsvenv'..."
+    venvcreate "${PYTHON:?}" authawsvenv || return
 
     local packages=(
       "boto3~=1.26.0"
--- a/.evergreen/auth_aws/activate-authawsvenv.sh
+++ b/.evergreen/auth_aws/activate-authawsvenv.sh
@@ -48,6 +48,7 @@ activate_authawsvenv() {
       deactivate || return 1 # Deactivation should never fail!
       return "$ret"
     }
+    echo "Creating virtual environment 'authawsvenv'... done."
   fi
 }
 
--- a/.evergreen/auth_aws/activate-authawsvenv.sh
+++ b/.evergreen/auth_aws/activate-authawsvenv.sh
@@ -34,8 +34,10 @@ activate_authawsvenv() {
     . ../find-python3.sh || return
 
     echo "Creating virtual environment 'authawsvenv'..."
-
-    venvcreate "$(find_python3 2>/dev/null)" authawsvenv || return
+    echo "Finding Python3 binary..."
+    PYTHON="$(find_python3 2>/dev/null)"
+    echo "Finding Python3 binary... done."
+    venvcreate $PYTHON authawsvenv || return
 
     local packages=(
       "boto3~=1.26.0"
--- a/.evergreen/auth_aws/activate-authawsvenv.sh
+++ b/.evergreen/auth_aws/activate-authawsvenv.sh
@@ -33,7 +33,9 @@ activate_authawsvenv() {
     # shellcheck source=.evergreen/find-python3.sh
     . ../find-python3.sh || return
 
-    venvcreate "$(find_python3)" authawsvenv || return
+    echo "Creating virtual environment 'authawsvenv'..."
+
+    venvcreate "$(find_python3 2>/dev/null)" authawsvenv || return
 
     local packages=(
       "boto3~=1.26.0"
@@ -41,21 +43,7 @@ activate_authawsvenv() {
       "pymongo[aws]~=4.0"
     )
 
-    if [[ "$OSTYPE" == cygwin && "$HOSTTYPE" == x86_64 ]]; then
-      local -r windows_os_name="$(systeminfo.exe /FO LIST | perl -lne 'print $1 if m/^OS Name:\s+(.*)$/' || true)"
-
-      if [[ "$windows_os_name" =~ 2016 ]]; then
-        # Avoid `RuntimeError: Could not determine home directory.` on
-        # windows-64-2016. See BUILD-16233.
-        python -m pip install -U "setuptools<65.0" || {
-          local -r ret="$?"
-          deactivate || return 1 # Deactivation should never fail!
-          return "$ret"
-        }
-      fi
-    fi
-
-    python -m pip install -U "${packages[@]}" || {
+    python -m pip install -q -U "${packages[@]}" || {
       local -r ret="$?"
       deactivate || return 1 # Deactivation should never fail!
       return "$ret"
--- a/.evergreen/find-python3.sh
+++ b/.evergreen/find-python3.sh
@@ -198,6 +198,8 @@ find_python3() (
   local -a bins=()
   local bin=""
 
+  trap "echo 'find_python3 failed!'" EXIT
+
   # The list of Python binaries to test for venv or virtualenv support.
   # The binaries are tested in the order of their position in the array.
   {
@@ -239,8 +241,6 @@ find_python3() (
     # Some environments trigger an unbound variable error if "${bins[@]}" is empty when used below.
     if (("${#bins[@]}" == 0)); then
       echo "Could not find any python3 binaries!"
-      echo "If you are suppressing stdout with 'find_python3 2>/dev/null' you may need to"
-      echo "remove it to aid in debugging."
       return 1
     fi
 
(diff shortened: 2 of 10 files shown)


.evergreen/auth_aws/activate-authawsvenv.sh
```diff
@@ -33,29 +33,17 @@ activate_authawsvenv() {
     # shellcheck source=.evergreen/find-python3.sh
     . ../find-python3.sh || return
 
-    venvcreate "$(find_python3)" authawsvenv || return
+    echo "Creating virtual environment 'authawsvenv'..."
```
eramongodb, line 36: A personal style/preference suggestion: add a corresponding message indicating when creation has successfully completed such that the output looks like:
```
... # other output
Creating virtual environment 'authawsvenv'...
Creating virtual environment 'authawsvenv'... done.
... # other output
```
This may slightly improve traceability if an EVG task ever fails due to this script ("creating" yet absence of "done") or if an error is not correctly handled ("done" is present despite error output).

This suggestion also applies to other instances of a new "Creating" message.

.evergreen/run-load-balancer.sh
```diff
@@ -77,9 +77,11 @@ EOF_HAPROXY_CONFIG
 }
 
 stop() {
-  echo "Stopping HAProxy..."
-  kill -USR1 $(cat $DRIVERS_TOOLS/haproxy.pid)
-  rm $DRIVERS_TOOLS/haproxy.conf $DRIVERS_TOOLS/haproxy.pid
+  if [ -f "$DRIVERS_TOOLS/haproxy.pid" ]; then
```
eramongodb, line 80:
```suggestion
  if [[ -f "$DRIVERS_TOOLS/haproxy.pid" ]]; then
```
Recommend using `[[` over `[` for modernization/consistency of Bash scripts.

eramongodb (approved):
Some minor suggestions; otherwise, LGTM.

I have a concern that if any errors should occur in the future (e.g. due to change in environment, updates to unpinned Python packages, etc.) that this may reduce the discoverability of the root cause due to hiding both informational _and_ error output (`2>/dev/null` doesn't distinguish them). However, given these scripts have now had fairly widespread usage experience, this may not be as much of a concern as I think it may be.

blink1073:
> I have a concern that if any errors should occur in the future (e.g. due to change in environment, updates to unpinned Python packages, etc.) that this may reduce the discoverability of the root cause due to hiding both informational and error output (2>/dev/null doesn't distinguish them). However, given these scripts have now had fairly widespread usage experience, this may not be as much of a concern as I think it may be.

Perhaps we should print a notice to stdout if it fails saying something like "If you are using find_python3 2>/dev_null you may need to remove it to debug this failure"?

.evergreen/auth_aws/activate-authawsvenv.sh
blink1073: ... done.   :)

eramongodb (commented):
> Perhaps we should print a notice to stdout if it fails saying something like "If you are using find_python3 2>/dev_null you may need to remove it to debug this failure"?

I don't think that will be necessary, as the `2>/dev/null` behavior is [already documented](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/a126caf6e4fd7d6e5ddf0740e01c4a256cb88ce6/.evergreen/find-python3.sh#L180). So long as task output makes it clear that `find_python3` is the cause of failure when it occurs, I think silencing the output is acceptable.

What I would like to avoid is a scenario where the task output looks like the following:

```
...
some unrelated script output
some unrelated script output
!task failed to due script error!
task cleanup output
```

where it is entirely unclear what the cause of error may be. I think its execution should at least be preceeded/followed by a status message (e.g. "Running find_python3..."), similar to what you've already done for the `activate-*venv.sh` scripts concerning venv creation. This would make the error easily tracable if an error occurs:

```
...
some unrelated script output
some unrelated script output
Running find_python3...
!task failed to due script error!
task cleanup output
```

.evergreen/find-python3.sh
```diff
@@ -239,6 +239,8 @@ find_python3() (
     # Some environments trigger an unbound variable error if "${bins[@]}" is empty when used below.
     if (("${#bins[@]}" == 0)); then
       echo "Could not find any python3 binaries!"
+      echo "If you are suppressing stdout with 'find_python3 2>/dev/null' you may need to"
```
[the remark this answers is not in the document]
blink1073: Ah, good point, I forgot these were in the capture blocks.

blink1073:
We're relying on stdout to be the python3 binary, so I used a `trap` instead.

.evergreen/find-python3.sh
```diff
@@ -198,6 +198,8 @@ find_python3() (
   local -a bins=()
   local bin=""
 
+  trap "echo 'find_python3 failed!'" EXIT
+
```
eramongodb, line 202:
```suggestion
```
I strongly discourage the use of `trap`.

Given `find-python3.sh` is designed to be sourced, this trap will be unconditionally set in the scope of the parent script sourcing this file. That means this trap command will only be executed when the _parent_ script exits (or whatever script(s) may be sourcing the parent, etc.), which is not the intent of this command. Furthermore, it may also silently override the caller's own trap commands. This script was deliberately designed to avoid any such "leaks" that may affect the user's scripting environment.

I recommend simply adding `echo` guards around calls to the `find_python3` function in files such as `start-orchestration.sh` or `activate-*venv.sh` rather than modifying `find-python3.sh`. It may be a bit repetitive/tedious but it is a far safer and simpler-to-maintain approach.

eramongodb (changes requested):

blink1073:
Hm, that doesn't work either.   What if, instead, we have a `-q` flag that we pass to suppress debug output?   Without the `-q` flag, you see the following stderr:

```
Finding python3 binary...
Found python3 binary <foo>.
```

blink1073:
Or else you have to opt in to the verbose output with `-v`.

blink1073:
@eramongodb thoughts on the flag approach?

eramongodb:
> thoughts on the flag approach?

That would require refactoring `find-python3.sh` to [conditionally pipe multiple blocks of diagnostic output](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/a126caf6e4fd7d6e5ddf0740e01c4a256cb88ce6/.evergreen/find-python3.sh#L236) (`{ ... } 1>&2`) to either `stderr` or `/dev/null` depending on the presence of a function argument, which I believe exceed the scope of this PR's initial goal.

I would like to [reiterate my recommendation](https://github.com/mongodb-labs/drivers-evergreen-tools/pull/343#discussion_r1309260817) to add echo guards around calls to `find_python3` instead, whose instances are already identified in this PR by the addition of the `2>/dev/null` redirects, e.g. for `start-orchestration.sh`:

```bash
echo "Finding Python 3 binary..."
PYTHON="$(find_python3 2>/dev/null)"
echo "Finding Python 3 binary... done."
```

blink1073:
Okay, fair, let's go for that

.evergreen/auth_aws/activate-authawsvenv.sh
```diff
@@ -33,33 +33,24 @@ activate_authawsvenv() {
...
+    echo "Finding Python3 binary..."
+    PYTHON="$(find_python3 2>/dev/null)"
```
eramongodb, line 38:
```suggestion
    PYTHON="$(find_python3 2>/dev/null)" || return
```
Suggest ensuring early return on error. This should avoid the following confusing output sequence:
```
...
Finding Python3 binary...
Finding Python3 binary... done.
script error: venvcreate requires a path to the virtual environment to create
```

which would be triggered by the subsequent call to `venvcreate` if `find_python3` failed (thus setting `PYTHON` to a null value -> null expansion in `$PYTHON` below -> only one argument).

In addition to `|| return`, wrapping `$PYTHON` in quotes (`"$PYTHON"`) would also avoid this error (emitting a slightly more helpful "venvcreate requires a Python binary to use for the virtual environment" error instead). Even better, the expansion can be changed to `"${PYTHON:?}"` to explicitly assert that the variable should not null when used at this point (although it may be redundant given the early return):

```bash
echo "Finding Python3 binary..."
PYTHON="$(find_python3 2>/dev/null)" || return
echo "Finding Python3 binary... done."
venvcreate "${PYTHON:?}" authawsvenv || return
```

.evergreen/auth_aws/activate-authawsvenv.sh
eramongodb: Suggest moving the `Creating virtual environment` echo after the `Finding Python3 binary... done.` echo to avoid nested output.

Before:
```
Creating virtual environment 'foo'...
Finding Python3 binary...
Finding Python3 binary... done.
Creating virtual environment 'foo'... done.
```

After:
```
Finding Python3 binary...
Finding Python3 binary... done.
Creating virtual environment 'foo'...
Creating virtual environment 'foo'... done.
```

eramongodb (approved):
Some feedback w.r.t. error handling and echo ordering; otherwise, LGTM.

.evergreen/auth_aws/activate-authawsvenv.sh
blink1073: Done

.evergreen/auth_aws/activate-authawsvenv.sh
blink1073: Done

PR #346, opened 2023-08-31
add unique function generation

baileympearson:
The setup and teardown atlas cluster scripts create unique clusters per-commit.  This can result in multiple CI tasks for the same commit attempting to create the same cluster.

This PR uses the `task_id` (a default evergreen expansion) to ensure that generated clusters are unique per-task.

Here's a patch from Node: https://spruce.mongodb.com/version/64f8955c3627e0397fdb50c0/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

Note that this patch is failing because of unrelated `npm` version issues (we're fixing this separately) but you can see the cluster successfully spins up and tears down.

### DIFF
--- a/.evergreen/atlas/setup-atlas-cluster.sh
+++ b/.evergreen/atlas/setup-atlas-cluster.sh
@@ -25,7 +25,7 @@ DRIVERS_ATLAS_GROUP_ID
 DRIVERS_ATLAS_LAMBDA_USER
 DRIVERS_ATLAS_LAMBDA_PASSWORD
 LAMBDA_STACK_NAME
-TASK_ID
+task_id
 )
 
 # Ensure that all variables required to run the test are set, otherwise throw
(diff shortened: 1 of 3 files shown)


.evergreen/atlas/setup-atlas-cluster.sh
```diff
@@ -25,6 +25,7 @@ DRIVERS_ATLAS_GROUP_ID
 DRIVERS_ATLAS_LAMBDA_USER
 DRIVERS_ATLAS_LAMBDA_PASSWORD
 LAMBDA_STACK_NAME
+task_id
```
baileympearson, line 28: although this doesn't follow the existing all-caps convention, I chose to use the same name as the default evergreen expansion because drivers can then rely on `add_expansions_to_env` to ensure the task id is present.

Node, for example, already uses `add_expansions_to_env` so no changes are necessary 😄

blink1073 (approved):
LGTM!

PR #347, opened 2023-09-01
DRIVERS-2585 Make the Secrets Access errors more user friendly

blink1073:
Instead of a traceback, they'll see something like:

```

ERROR: The SSO session associated with this profile has expired or is otherwise invalid. To refresh this SSO session run aws sso login with the corresponding profile.

```


NoahStapp (approved):
LGTM!

PR #354, opened 2023-09-29
PYTHON-3942 Enable AzureKMS through AWS Secrets Manager

blink1073:
- Allows AzureKMS credentials to be pulled from AWS Secrets Manager using the instance profile credentials on the EVG hosts.
- Removes the `Key Vault Crypto User` role as part of teardown, if `AZUREKMS_SCOPE` is set.  We were adding orphaned roles to the key vault.
- Passing build: https://spruce.mongodb.com/task/mongo_python_driver_testazurekms_variant_testazurekms_task_patch_9b6f2e18cfcdf56ad2afc988246060c4d20e11b8_65160838e3c331a022edc469_23_09_28_23_12_00/logs?execution=0
- Passing build with Go Driver using this branch (to show there are no regressions): https://spruce.mongodb.com/task/mongo_go_driver_testazurekms_variant_testazurekms_task_patch_60ded84e486dda6eea4f446d642571489a3ed3d0_6516199557e85ac21791c6f2_23_09_29_00_26_08/logs?execution=0
- Needed for https://github.com/mongodb/mongo-python-driver/pull/1381


kevinAlbs (approved):
Updated scripts [pass with C driver](https://spruce.mongodb.com/task/mongo_c_driver_testazurekms_variant_testazurekms_task_patch_8a0e9c28b32724c5eb1493bc100fddb505433671_651721463e8e86808cd52f13_23_09_29_19_11_10/logs?execution=0).

Removal of orphaned roles is much appreciated. LGTM.

.evergreen/auth_aws/setup_secrets.py
```diff
@@ -6,19 +6,34 @@
 import json
 import os
 import sys
+import uuid
 
 import boto3
 import botocore.exceptions
 
+AWS_ROLE_ARN = "arn:aws:iam::857654397073:role/drivers-test-secrets-role"
```
NoahStapp, line 14: Could we set this in an environment variable? Not a huge fan of setting even non-secret credentials inside a script.

NoahStapp (changes requested):

.evergreen/auth_aws/setup_secrets.py
blink1073: There are other places we hard-code ARNs or links in this repo.  The intent here was that it should be entirely self-contained without drivers needing to manually update EVG project config, which is part of the reason for having this script in general.

NoahStapp: I'd prefer we migrate away form hard-coding ARNs in general, but that's a different discussion entirely. In the context of the existing repo, makes sense.

NoahStapp (approved):

PR #352, opened 2023-09-25
GODRIVER-2924 Add docker file support

blink1073:
Used in conjunction with https://github.com/mongodb/mongo-go-driver/pull/1381 to enable a local topology to be run in an isolated environment, or a set of drivers tests to be run in an isolated environment. 

Updates OIDC to use the new Dockerfile as a base.

Adds tests for docker and OIDC in EVG (will be added to default PR tasks).

PR #357, opened 2023-10-02
Address build failures due to Python 3 binary lookup

eramongodb:
Partially reverts some changes made in https://github.com/mongodb-labs/drivers-evergreen-tools/pull/352 concerning Python 3 binary lookup (motivated by [a stray explicit call to `python`](https://github.com/mongodb-labs/drivers-evergreen-tools/commit/f825f86fae3e5eecfa5da05a510ec5962fc7de88#diff-5eb412fc2cdeb6fab58a38e622949bedbda012b8bcb61e5e6daec62e5168af87L101)) to address build failures on some EVG distros where neither `python` nor `python3` are available in default binary paths.

The `Using Python3 binary: $PYTHON` case added in https://github.com/mongodb-labs/drivers-evergreen-tools/pull/352 is preserved to avoid a redundant call to `find_python3` when `start-orchestration.sh` is invoked via `run-orchestration.sh` (which already called `find_python3`). This _might_ still break EVG builds that happen to set the `PYTHON` environment variable before invoking `start-orchestration.sh` directly (instead of through `run-orchestration.sh`) if `PYTHON` is set to a Python binary that is not 3.0+ or does not support venv. If that is the case, this too can be reverted to the prior unconditional call to `find_python3`.

### DIFF
--- a/.evergreen/run-orchestration.sh
+++ b/.evergreen/run-orchestration.sh
@@ -30,9 +30,10 @@ DIR=$(dirname $0)
 # Functions to fetch MongoDB binaries.
 . $DIR/download-mongodb.sh
 
-. $DIR/find-python3.sh
+# To continue supporting `sh run-orchestration.sh` for backwards-compatibility,
+# explicitly invoke Bash as a subshell here when running `find_python3`.
 echo "Finding Python3 binary..."
-PYTHON="$(find_python3 2>/dev/null)"
+PYTHON="$(bash -c ". $DIR/find-python3.sh && find_python3 2>/dev/null")"
 echo "Finding Python3 binary... done."
 
 get_distro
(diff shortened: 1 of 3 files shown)


blink1073:
Test failure: `find-python3.sh must be run in a Bash shell!`

blink1073 (approved):
LGTM, thanks!

jmikola (approved):

eramongodb:
Although the GitHub Actions config was updated to use `shell: bash` for consistency, the `run-orchestration.sh` script itself still uses `#!/bin/sh` for backwards-compatibility, as many DET users still run `sh run-orchestration.sh`. Making `run-orchestration.sh` require Bash would be a change far beyond the scope of this PR; therefore, a Bash subshell is used to run `find_python3` instead. Documented this context with a comment prior to running `find_python3`.

blink1073:
Test failure is unrelated, was fixed in #356

PR #360, opened 2023-10-10
DRIVERS-2416 Add support for a second Azure client

blink1073:
This infrastructure will be needed to support testing drivers with multiple clients on Azure.  I tested it locally by running `create_and_update_vm.sh` and ensuring that the self-test passes.

### DIFF
--- a/.evergreen/auth_oidc/azure/test.py
+++ b/.evergreen/auth_oidc/azure/test.py
@@ -1,4 +1,5 @@
 from pymongo import MongoClient
+from functools import partial
 import os
 import json
 from urllib.request import urlopen, Request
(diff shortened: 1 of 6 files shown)


.evergreen/auth_oidc/azure/test.py
```diff
@@ -8,11 +8,13 @@
 _AUTH_MAP["MONGODB-OIDC"] = _authenticate_oidc
 
 app_id = os.environ['AZUREOIDC_CLIENTID']
+client_id = os.environ['AZUREOIDC_TOKENCLIENT']
 
 def callback(client_info, server_info):
```
matthewdale, line 13: Optional: The side effect of setting `client_id` on `callback` is fairly confusing. Is there any way to encode the new `client_id` requirement in the `callback` function, or in a `callback` creator function?

E.g.
```python
def callback_for(app_id, client_id):
    def callback(client_info, server_info):
        ...

    return callback
```

matthewdale (approved):
Looks good 👍

PR #366, opened 2023-10-19
DRIVERS-2328: Skip installing the legacy shell by default

alcaeus:
DRIVERS-2328

This PR removes the current `SKIP_LEGACY_SHELL` environment variable and introduces a new `INSTALL_LEGACY_SHELL` variable to essentially flip the default behaviour.

[Recent work in the Go driver](https://github.com/mongodb/mongo-go-driver/pull/1429) provides a template for drivers to remove usage of the legacy shell. The "Downstream Changes Summary" of the above ticket reads as follows:
> The legacy shell will no longer be installed in the near future. Drivers are encouraged to remove all usages of the legacy shell in their drivers, with this [Go driver PR](https://github.com/mongodb/mongo-go-driver/pull/1429) serving as inspiration.
>
> All drivers that require the use of the legacy library are required to define the `INSTALL_LEGACY_SHELL` environment variable to a non-zero value when running MongoDB installation (either by running `run-orchestration.sh` or by manually invoking `download_and_extract()` from `download-mongoldb.sh`). This variable replaces the current `SKIP_LEGACY_SHELL` variable which will no longer be respected.

If this approach sounds good, I would send the drivers ticket to implementation and add a timeline for this change to be merged. I would propose one month for this, as the change would allow drivers to keep using the legacy shell by adding `INSTALL_LEGACY_SHELL=1` to appropriate places before taking the time to remove legacy shell usage.


blink1073 (approved):
LGTM!

jmikola:
@alcaeus: Should this still be in a draft state?

PR #383, opened 2023-12-19
 DRIVERS-2416 Azure OIDC updates

blink1073:
- Add only one managed identity to the VM
- Use the object id of the managed identity as the `username` instead of as an auth mechanism property

Tested with https://github.com/mongodb/mongo-python-driver/pull/1443


matthewdale (approved):
Looks good! 👍

durran (approved):

PR #384, opened 2024-01-08
Normalize Script Path Usage

blink1073:
#383 caused a regression in the way paths are handled in `aws_setup.sh`:

`mv: 'secrets-export.sh' and '/data/mci/1e7ceedf7f78683b5034736f250f800b/drivers-tools/.evergreen/auth_aws/secrets-export.sh' are the same file`

https://spruce.mongodb.com/task/mongo_go_driver_v1_aws_auth_test__version~4.4_os_aws_auth~ubuntu2004_64_go_1_20_aws_auth_test_1415137a1a3cbf6886ad7d6a4a576c2d05f4f83c_24_01_05_17_26_12/logs?execution=0

This PR adjusts path handling so that the path to the script is always absolute across platforms, and finds the appropriate parent DRIVERS_TOOLS directory, using a new `handle-paths.sh` script.

Here is a passing build for [pymongo](https://spruce.mongodb.com/version/659b09aed6d80ad0c4b2c88c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC
) across all build variants (except Atlas Data Lake, which is unrelated).

### DIFF
--- a/.evergreen/atlas/setup-atlas-cluster.sh
+++ b/.evergreen/atlas/setup-atlas-cluster.sh
@@ -36,7 +36,7 @@ for VARNAME in ${VARLIST[*]}; do
 done
 
 # Set up the common variables.
-DIR="$(dirname "${BASH_SOURCE[0]}")"
+DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 . $DIR/setup-variables.sh
 
 # The cluster server version.
--- a/.evergreen/atlas/setup-atlas-cluster.sh
+++ b/.evergreen/atlas/setup-atlas-cluster.sh
@@ -36,7 +36,7 @@ for VARNAME in ${VARLIST[*]}; do
 done
 
 # Set up the common variables.
-DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+DIR=$(dirname ${BASH_SOURCE:-$0})
 . $DIR/../handle-paths.sh
 . $DIR/setup-variables.sh
 
--- a/.evergreen/atlas/setup-atlas-cluster.sh
+++ b/.evergreen/atlas/setup-atlas-cluster.sh
@@ -36,9 +36,9 @@ for VARNAME in ${VARLIST[*]}; do
 done
 
 # Set up the common variables.
-DIR=$(dirname ${BASH_SOURCE:-$0})
-. $DIR/../handle-paths.sh
-. $DIR/setup-variables.sh
+SCRIPT_DIR=$(dirname ${BASH_SOURCE[0]})
+. $SCRIPT_DIR/../handle-paths.sh
+. $SCRIPT_DIR/setup-variables.sh
 
 # The cluster server version.
 VERSION="${MONGODB_VERSION:-6.0}"
--- a/.evergreen/atlas/setup-atlas-cluster.sh
+++ b/.evergreen/atlas/setup-atlas-cluster.sh
@@ -37,6 +37,7 @@ done
 
 # Set up the common variables.
 DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+. $DIR/../handle-paths.sh
 . $DIR/setup-variables.sh
 
 # The cluster server version.
--- a/.evergreen/auth_oidc/azure/create-and-setup-vm.sh
+++ b/.evergreen/auth_oidc/azure/create-and-setup-vm.sh
@@ -11,7 +11,7 @@ if [ -z "${AZUREOIDC_VMNAME_PREFIX:-}" ]; then
 fi
 
 DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
-DRIVERS_TOOLS=$(dirname $(dirname $(dirname $DIR)))
+. $DIR/../../handle-paths.sh
 
 # Set defaults.
 BASE_PATH="$DRIVERS_TOOLS/.evergreen/auth_oidc"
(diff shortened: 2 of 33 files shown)


.evergreen/handle-paths.sh
```diff
@@ -0,0 +1,37 @@
...
+  DIR="$( cd -- "$DIR" &> /dev/null && pwd )"
+fi
+if [ "Windows_NT" = "${OS:-}" ]; then # Magic variable in cygwin
+  DIR=$(cygpath -m $DIR)
+fi
+
+# Find the DRIVERS_TOOLS by walking up the folder tree until there
+# is a .evergreen folder in the same directory.
+if [ -z "${DRIVERS_TOOLS:-}" ]; then
+  DRIVER_TOOLS=$(dirname $DIR)
```
alcaeus, line 30: This should be `DRIVERS_TOOLS`:
```suggestion
  DRIVERS_TOOLS=$(dirname $DIR)
```

.evergreen/handle-paths.sh
```diff
@@ -0,0 +1,37 @@
...
+  while 1
+  do
+    if [ -d "$DRIVERS_TOOLS/.evergreen" ]; then 
+      break 
+    fi
+  done
```
alcaeus, line 36: I'm a little confused. Where in this loop does the "walking up the folder tree" happen? From what I can see, we check if `$DRIVERS_TOOLS/.evergreen` is a directory and break if it is. In the case it isn't, we repeat the loop without changing `$DRIVERS_TOOLS` at all, so unless `$DRIVERS_TOOLS` is set to the correct directory or `dirname $DIR` returns the correct directory, this will be an endless loop. Did I miss something here?

.evergreen/handle-paths.sh
blink1073: Fixed

.evergreen/handle-paths.sh
blink1073: No, I did, fixed.  :)

.evergreen/atlas/setup-atlas-cluster.sh
```diff
@@ -36,7 +36,8 @@ for VARNAME in ${VARLIST[*]}; do
 done
 
 # Set up the common variables.
-DIR="$(dirname "${BASH_SOURCE[0]}")"
+DIR=$(dirname ${BASH_SOURCE:-$0})
```
alcaeus, line 39: I actually had to go and double-check why this works. For those who also didn't know, bash allows one to reference array element `0` of an array using scalar notation. For more details, consider the first "Optional Reading" section in [this StackOverflow answer](https://stackoverflow.com/a/35006505).

.evergreen/auth_oidc/azure/create-and-setup-vm.sh
```diff
@@ -10,9 +10,11 @@ if [ -z "${AZUREOIDC_VMNAME_PREFIX:-}" ]; then
     exit 1
 fi
 
+DIR=$(dirname ${BASH_SOURCE:-$0})
+. $DIR/../../handle-paths.sh
+
 # Set defaults.
-AZUREOIDC_DRIVERS_TOOLS=${AZUREOIDC_DRIVERS_TOOLS:-$DRIVERS_TOOLS}
```
alcaeus, line 14: Was there a reason why a separate variable was introduced here? I don't think it is necessary, just doing my due diligence here.

alcaeus (approved):
Thanks for cleaning this up. I'm amazed how many different ways we had to get a specific absolute path.

I ran a patch build with the PHP driver, which didn't yield any failures. I should also mention that we don't use drivers-evergreen-tools on Windows, so if you haven't already checked that you may want to ask somebody to test these changes on Windows.

.evergreen/atlas/setup-atlas-cluster.sh
blink1073: I actually had used that answer.  :).  I added a link.

blink1073:
I ran the python tests on Windows.  Thanks for the review!

.evergreen/handle-paths.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
+#
+# This script will handle the correct cross-platform absolute
+# paths for a script directory and DRIVERS_TOOLS.  
+# It is meant to be invoked as follows:
+#
+# DIR=$(dirname ${BASH_SOURCE:-$0})
```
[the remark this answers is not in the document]
blink1073: Okay, what I'm hearing is that we need to split this into two scripts: one to get the absolute script dir where needed, and one to get just DRIVERS_TOOLS.  Both scripts will use echo to get the value.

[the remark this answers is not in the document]
blink1073: I changed it to use `SCRIPT_DIR` and use `BASH_SOURCE[0]` where we know we require `bash`.  I also kicked the python builds and the failing C builds.

.evergreen/handle-paths.sh
```diff
@@ -0,0 +1,38 @@
...
+fi
+if [ "Windows_NT" = "${OS:-}" ]; then # Magic variable in cygwin
+  SCRIPT_DIR=$(cygpath -m $SCRIPT_DIR)
+fi
+
+# Find the DRIVERS_TOOLS by walking up the folder tree until there
+# is a .evergreen folder in the same directory.
+if [ -z "${DRIVERS_TOOLS:-}" ]; then
+  DRIVERS_TOOLS=$(dirname $SCRIPT_DIR)
+  while 1
```
kevinAlbs:
```suggestion
  while true
```

This fixes an [observed failure](https://spruce.mongodb.com/task/mongo_c_driver_testazurekms_variant_testazurekms_task_patch_6f4d492aacc31bcfd81e0244456b58de2ef9c84c_659c0c933e8e86c1a4c04154_24_01_08_14_54_13/logs?execution=0) in C driver tasks: `1: command not found`

blink1073: Fixed in the previous commit.  :)

kevinAlbs (approved):

PR #388, opened 2024-01-12
Add missing executable flags on scripts

alcaeus:
This adds the x flag to a few shell scripts that previously didn't have it so they don't have to be invoked through a shell but can rely on the shebang.


blink1073 (approved):
Thank you!

PR #399, opened 2024-01-29
DRIVERS-2818 Apply auto-formatters

blink1073:
Used https://github.com/mongodb-labs/drivers-atlas-testing/blob/master/.pre-commit-config.yaml and https://github.com/mongodb-labs/drivers-atlas-testing/blob/master/pyproject.toml (which we will change to just `ruff.toml` since there is no top level project).


blink1073:
Testing in Go with https://spruce.mongodb.com/version/65b8285ec9ec44494945aee9/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC.

blink1073:
I'll pick this back up at some other time.

PR #400, opened 2024-02-01
DRIVERS-2384 Allow extra args to sam build

abr-egn:
The motivating use case here is Rust, which requires the `--beta-features` flag.


.evergreen/aws_lambda/run-deployed-lambda-aws-tests.sh
```diff
@@ -84,7 +88,7 @@ trap cleanup EXIT SIGHUP
 
 cd "${TEST_LAMBDA_DIRECTORY}"
 
-sam build
+sam build ${SAM_BUILD_ARGS}
```
blink1073, line 91:
```suggestion
sam build ${SAM_BUILD_ARGS:-}
```

blink1073 (approved):
Thanks!

PR #402, opened 2024-02-05
DRIVERS-2384 Skip deployment confirmation

abr-egn:
Honestly a little confused as to how other drivers setups are working; without this, the Rust evergreen run fails when it hits the `Deploy this changeset? [y/N]` prompt.


blink1073 (approved):
Thank you!

PR #407, opened 2024-02-21
DRIVERS-2843 Add Convenience Setup and Teardown Scripts

blink1073:


blink1073:
evergreen retry

blink1073:
evergreen retry

.evergreen/atlas/teardown.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+# Handle proper teardown of Atlas.
```
eramongodb, line 2: imo this documentation comment is unnecessary: the name of the script and its parent directory clearly describes the purpose of this script. Suggest its removal from this + similar scripts.

.evergreen/config.yml
```diff
@@ -17,6 +17,12 @@ command_type: system
 # That roughly accounts for variable system performance for various buildvariants
 exec_timeout_secs: 1800 # 6 minutes is the longest we'll ever run
 
+# Ensure that setup and teardown is working as expected.
+pre_error_fails_task: true
+ppre_timeout_secs: 1800 # 5 minutes
+post_error_fails_task: true
+post_timeout_secs: 1800 # 5 minutes
```
eramongodb, line 24: Confirming, are _all_ the pre and post commands currently defined and used by this EVG config prepared for these fields to be set (failure-resilience)?

.evergreen/atlas/teardown.sh
```diff
@@ -0,0 +1,11 @@
...
+
+set -o errexit  # Exit the script with error if any of the commands fail
```
eramongodb, line 4: imo this documentation comment is unnecessary: it provides no additional context or information beyond what is immediately apparant ("yes, that is exactly what `errexit` does"). Suggest its removal from this + similar scripts.

eramongodb (commented):
What motivates the proposed changes in this PR? If it is simply the convenience of a global setup/teardown procedure via pre/post commands, I am against the proposed changes.

> Using this script in a `post:` task can remove the need for `task_groups` to ensure that services are properly torn down.

On the contrary, I believe we should be actively _discouraging_ use of global pre/post commands, which are executed regardless of the relevance of the task being executed.

The set of tasks that actually depend on a given set of setup/teardown scripts is likely to be a small portion of the overall set of scheduled tasks per build for any given EVG configuration. Isolating relevant setup/teardown commands to the set of tasks that actually requires them via task groups applies [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns) and improves the resilience of EVG configurations. e.g. consider the impact of just _one_ of the setup or teardown commands failing (i.e. due to broken credentials, network errors, or any of the other miscellaneous reasons historically encountered by DET users) on the _entire_ EVG build, regardless whether any given task even cares about the related creds/services/etc. The proposed changes in this PR seem to work against separation of concerns by encouraging the unconditional coupling of _all_ setup/teardown commands to single convenience scripts run in global pre/post commands regardless of their relevance to any particular task (excluding task groups, of course).

I would not mind the introduction of helper functions/scripts that unify the use and/or implementation of the various setup/teardown scripts (i.e. venv-utils.sh as now used by `activate-*.sh` scripts for Python virtual environment handling), but the introduction of global scripts that "setup/teardown _all_ the things" does not seem like a favorable direction to pursue.

blink1073:
Fair points, here is my proposal:

- Each directory is encouraged to have a `setup-test.sh` and `teardown-test.sh` file to make things uniform.
- The top-level `setup-test.sh` and `teardown-test.sh` only applies to the top level folder (basically as it is now, but removing the handling of the sub-folders.

eramongodb:
> Each directory is encouraged to have a setup-test.sh and teardown-test.sh file to make things uniform.

I think establishing a consistent scheme like this is preferable.

> The top-level setup-test.sh and teardown-test.sh only applies to the top level folder (basically as it is now, but removing the handling of the sub-folders).

It remains unclear to me what the scope of responsibility of the "top-level" setup/teardown scripts is supposed to cover (mongo-orchestration? load balancers? docker containers? atlas proxy?). Each of these should arguably be subdirectories, but I suspect their relocation is not feasible due to compatibility concerns. So long as the scope is clear and well-documented (to avoid potential scope creep of scripts' responsibilities), I think that is acceptable.

That being said, I think more effort can be made to reduce the scope of responsibility to related components. e.g. `ABSOLUTE_PATH_REPLACEMENT_TOKEN` handling is [already being done in run-orchestration.sh](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/ca29dc0cca7f7fcb6de012e53b918c0e1958fec8/.evergreen/run-orchestration.sh#L108); can/should the handling of the `$MONGO_ORCHESTRATION_HOME/orchestration.config` and `${MONGO_ORCHESTRATION_HOME}/lib/client.pem` files be handled there as well (are they really a "top-level" concern)?

.evergreen/atlas/teardown.sh
blink1073: Done

.evergreen/config.yml
blink1073: That is the intent, I'll double-check by scheduling all of the tasks.

.evergreen/atlas/teardown.sh
blink1073: Removed

blink1073:
I believe I've addressed your feedback.

blink1073:
I see some failures, but my internet is not cooperating, I'll return to this tomorrow.

.evergreen/handle-paths.sh
```diff
@@ -36,3 +36,7 @@ if [ -z "${DRIVERS_TOOLS:-}" ]; then
     DRIVERS_TOOLS=$(dirname $DRIVERS_TOOLS)
   done
 fi
+
+if [ "Windows_NT" = "${OS:-}" ]; then # Magic variable in cygwin
```
eramongodb, line 40: `$OS` is a highly overloaded variable used by various scripts both in DET and by Drivers' own scripts. Recommend using `$OSTYPE` instead (Bash variable, less likely to be overwritten by user scripts):

```bash
if [[ "${OSTYPE:?}" = cygwin ]]; then
```

eramongodb (approved):
Miscellaneous suggestions and feedback. Otherwise, LGTM.

README.md
```diff
@@ -103,6 +103,24 @@ To run an individual hook like `shellcheck` manually, run:
 pre-commit run --all-files shellcheck
 ```
 
+## Setup and Teardown
+
+For convenience, there are two scripts that can be used for setup and teardown of assets and services
+used by `drivers-evergreen-tools`.
+
+`$DRIVERS_TOOLS/.evergreen/setup-test.sh` will handle common setup actions that have previously
```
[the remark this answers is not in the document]
blink1073: Done

.gitignore
```diff
@@ -117,4 +117,5 @@ secrets-export.sh
 token_file.txt
 pykmip.log.*
 nohup.out
+test_logs.tar.gz
```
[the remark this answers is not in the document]
blink1073: Done

.evergreen/teardown-test.sh
```diff
@@ -0,0 +1,53 @@
...
+
+set -eux
+
+SCRIPT_DIR=$(dirname ${BASH_SOURCE[0]})
+. $SCRIPT_DIR/handle-paths.sh
+
+pushd $SCRIPT_DIR
+
+# Stop orchestration if it is running.
+if [ -f "${MONGO_ORCHESTRATION_HOME}/server.log" ]; then
```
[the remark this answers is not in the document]
blink1073: Done

.evergreen/run-atlas-proxy.sh
```diff
@@ -80,7 +80,7 @@ AP_END=$(date +%s)
 # Write results file
 DL_ELAPSED=$(expr $DL_END - $DL_START)
 AP_ELAPSED=$(expr $AP_END - $AP_START)
-cat <<EOT >> $DRIVERS_TOOLS/results.json
+cat <<EOT > $DRIVERS_TOOLS/results.json
```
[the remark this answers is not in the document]
blink1073: Done

.evergreen/handle-paths.sh
blink1073: Replaced all usages

.evergreen/teardown-test.sh
```diff
@@ -0,0 +1,53 @@
...
+# Clean up docker.
+DOCKER=$(command -v docker) || true
+if [ -n "$DOCKER" ]; then
+    # Kill all containers.
+    docker rm $(docker ps -a -q) &> /dev/null || true
+    # Remove all images.
+    docker rmi -f $(docker images -a -q) &> /dev/null || true
+fi
+
+# Move all child log files into $DRIVERS_TOOLS/.evergreen/test_logs.tar.gz.
```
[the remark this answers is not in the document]
blink1073: I tried that, but the resulting file is overly nested and has the hidden `.evergreen` file in it, which you have to remember the keyboard shortcut for in Finder.

image

.evergreen/handle-paths.sh
blink1073: I ended up using `if [[ "$(uname -s)" == CYGWIN* ]]; then` consistently since it works from sh as well.

eramongodb (changes requested):

.evergreen/handle-paths.sh
blink1073: I changing to `bash` would need to be a separate DRIVERS ticket, outside the scope of this change.

blink1073: Let's assume for now that only `run-orchestration.sh` and `download-mongodb.sh` need to be able to be run from `sh`, and inline `handle-paths.sh` into `run-orchestration.sh`.  I'll open up a ticket to ask drivers to audit their use of `sh` for drivers-tools scripts.

blink1073: On second thought, `handle-paths.sh` was too much to inline.  I opened https://jira.mongodb.org/browse/DRIVERS-2852.

eramongodb: Note, `sh` _may_ invoke a `bash` script via `bash path/to/script.sh` so long as it does not require any variables to be modified/exported/etc. by `script.sh`. Similarly, `sh` may be invoked from a `bash` script via `sh path/to/script.sh` with similar restrictions. Since `handle-paths.sh` is designed to modify existing variables and is being sourced from so many scripts, I support changing the shebang to `sh` to indicate sh-compatibility.

.evergreen/stop-orchestration.sh
```diff
@@ -1,11 +1,19 @@
 #!/bin/sh
 set -o errexit  # Exit the script with error if any of the commands fail
 
+# Fix orchestration path.
+if [[ "$(uname -s)" == CYGWIN* ]]; then
+  MONGO_ORCHESTRATION_HOME=$(cygpath -m $MONGO_ORCHESTRATION_HOME)
+fi
```
eramongodb, line 7: Confirming, what prompted these additions of `cygpath -m $MONGO_ORCHESTRATION_HOME` to become necessary? (I don't believe they were necessary prior to this PR) Can the root cause be identified and reverted instead of relying on calls to `cygpath`?

eramongodb (approved):
Aside from the `cygpath -m $MONGO_ORCHESTRATION_HOME` issue, LGTM.

.evergreen/stop-orchestration.sh
blink1073: The idea was to move the logic out of EVG config.  My follow up idea was for drivers to write a config env file, and then we normalize that file.   For now, we can move the logic back into EVG config.

blink1073: Hmm, actually the issue was that "OSTYPE" was not defined on Windows in `shell.exec`, even though the shell was set to `bash`.  So I had to change that to use `uname` as well.

PR #409, opened 2024-02-26
Clean up EVG tasks and remove unused scripts

blink1073:
- MongoDB 7.0.x is no longer supported on Debian 10
- Use a "pr" tag for tasks that should run on a PR
- Remove unused `csfle` scripts (replaced by hyphenated versions).


alcaeus (approved):
Changes LGTM. Do we need downstream tickets to let people know that the old scripts were removed?

blink1073:
>  Do we need downstream tickets to let people know that the old scripts were removed?

Only Python and Go were using them, already addressed.  :smile:

PR #415, opened 2024-03-06
DRIVERS-2851 Fix Handling of Atlas Cluster Creation

blink1073:
- Enforce status code when creating cluster.
- Remove the need to pass in `execution` and `task_id` by storing the cluster name.
- Make Atlas standalone and not tied to Lambda, since it is used in other contexts.
- Make the Atlas base URI configurable so we can change it to non-prod.
- Add Atlas Test.

Tested with the Go Driver [here](https://spruce.mongodb.com/version/65e7dc5a562343a977ef45c6/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

### DIFF
--- a/.evergreen/atlas/README.md
+++ b/.evergreen/atlas/README.md
@@ -22,22 +22,25 @@ An example task group running on a Linux EVG host might look like:
       params:
         binary: bash
         env:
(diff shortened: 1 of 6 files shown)


blink1073:
cc @durran

durran (approved):

PR #431, opened 2024-04-18
DRIVERS-1541 mock server for retry testing

adriandole:
Functions mocked:
- GCP auth, encrypt, decrypt
- Azure auth, encrypt, decrypt
- AWS encrypt, decrypt

The mock server fails every other KMS decrypt request with a 429 error.

### DIFF
--- a/.evergreen/csfle/kms_failpoint_server.py
+++ b/.evergreen/csfle/kms_failpoint_server.py
@@ -132,7 +132,7 @@ class Handler(http.server.BaseHTTPRequestHandler):
                 self._decrypt_fail()
                 return
             return self._send_json({"plaintext": base64.b64encode(fake_plaintext.encode()).decode()})
-        # Azure decrypt path: /keys/{key-name}/{key-version}/unwrapkey  
+        # Azure decrypt path: /keys/{key-name}/{key-version}/unwrapkey
         elif path.match("*unwrapkey"):
             if remaining_decrypt_fails > 0:
                 self._decrypt_fail()


.evergreen/csfle/activate-kmstlsvenv.sh
```diff
@@ -44,6 +44,7 @@ activate_kmstlsvenv() {
       "boto3~=1.19.0"
       "git+https://github.com/kevinAlbs/PyKMIP.git@DRIVERS-2659" # Add work around for DRIVERS-2659
       "sqlalchemy<2.0.0" # sqlalchemy.exc.InvalidRequestError: Implicitly combining column managed_objects.uid with column crypto_objects.uid under attribute 'unique_identifier'.
+      "flask>=3.0"
```
kevinAlbs, line 47: Instead of Flask, consider using the built-in `http.server`. This may enable testing network errors.

Here is an example: https://gist.github.com/kevinAlbs/3b3b84359bb0e350bf494597bcff0167

It also adds a `/set_failpoint` endpoint to tell the server to fail the next decrypt request. That may enable driver tests to more easily control how/when the test is expected to fail.

.evergreen/csfle/kms_failpoint_server.py
```diff
@@ -0,0 +1,155 @@
...
+$ curl -X POST localhost:3000
+curl: (52) Empty reply from server
+
+$ curl -X POST localhost:3000
+{"message": "ok"}
+
+"""
+
+# A new instance of Hander is created for every request, so these have to be global variables
+failpoint_type = None
```
kevinAlbs, line 26: Suggest removing `failpoint_type` as a global. It is only used to determine which failpoint to set in `do_GET`.

kevinAlbs (approved):
LGTM with minor feedback.

.evergreen/csfle/kms_failpoint_server.py
```diff
@@ -0,0 +1,155 @@
...
+fake_plaintext = 'b' * 96
+
+class HTTPServerWithTLS(http.server.HTTPServer):
+    def __init__(self, server_address, Handler):
+        super().__init__(server_address, Handler)
+        self.socket = ssl.wrap_socket(
+            self.socket,
+            server_side=True,
+            certfile='../x509gen/server.pem',
+            ca_certs='../x509gen/ca.pem',
```
vector-of-bool, line 40: This relies on the cwd being correct, I assume? I recommend either parameterizing this or getting the location of these files relative to the `__file__` global.

.evergreen/csfle/kms_failpoint_server.py
```diff
@@ -0,0 +1,155 @@
...
+        global remaining_decrypt_fails
+        remaining_decrypt_fails -= 1
+        self.send_response(429)
+    
+    def do_GET(self):
+        global failpoint_type
+        global remaining_decrypt_fails
+        global remaining_network_fails
+        parts = urllib.parse.urlsplit(self.path)
+        path = parts[2]
```
vector-of-bool, line 74: Recommend using `pathlib.PurePosixPath` for robust path handling here

.evergreen/csfle/kms_failpoint_server.py
```diff
@@ -0,0 +1,155 @@
...
+        self.send_header("Content-Length", len(msg))
+        self.end_headers()
+        self.wfile.write(msg)
+
+    def _decrypt_fail(self):
+        global remaining_decrypt_fails
+        remaining_decrypt_fails -= 1
+        self.send_response(429)
+    
+    def do_GET(self):
```
vector-of-bool, line 69: I'm iffy on using GET to modify the server state. Could this go into a PUT/POST handler?

.evergreen/csfle/kms_failpoint_server.py
```diff
@@ -0,0 +1,155 @@
+import argparse
+import http.server
+import json
+import urllib
+import ssl
+import base64
+
+"""
```
vector-of-bool, line 8: Minor: move this doc-string above the imports.

ShaneHarvey:
Is this ready for review?

PR #434, opened 2024-04-23
Fix serverless cluster setup

alcaeus:
#433 removed up usage of `jq` in atlas-utils.sh, but in doing so introduced a small typo which led to an error when parsing the response. There was also an additional usage of `jq` left in setup-cluster.sh, which I removed with analogous code. I've tested these changes and can confirm that they fix PHP's serverless tests on Debian 11.

### DIFF
--- a/.evergreen/serverless/create-instance.sh
+++ b/.evergreen/serverless/create-instance.sh
@@ -25,6 +25,7 @@ set +o xtrace # Disable xtrace to ensure credentials aren't leaked
 CURRENT_DIR=$(pwd)
 SCRIPT_DIR=$(dirname ${BASH_SOURCE[0]})
 . $SCRIPT_DIR/../handle-paths.sh
+. $SCRIPT_DIR/../find-python3.sh
 pushd $SCRIPT_DIR
 
 # Load the secrets file if it exists.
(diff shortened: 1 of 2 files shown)


.evergreen/serverless/create-instance.sh
```diff
@@ -105,12 +106,12 @@ if [ $SERVERLESS_URI = "null" ]; then
...
 
-SERVERLESS_MONGODB_VERSION=$(curl -sS \
+RESP=$(curl -sS \
   --digest -u "${ATLAS_PUBLIC_API_KEY}:${ATLAS_PRIVATE_API_KEY}" \
   -X GET \
-  "${ATLAS_BASE_URL}/groups/${ATLAS_GROUP_ID}/serverless/${DEPLOYMENT_NAME}" \
-  | jq -r '.mongoDBVersion'
+  "${ATLAS_BASE_URL}/groups/${ATLAS_GROUP_ID}/serverless/${DEPLOYMENT_NAME}"
 );
+SERVERLESS_MONGODB_VERSION=$($PYTHON -c "import json;d=json.loads('${RESP}');print(d['mongoDBVersion'])")
```
blink1073, line 114: I think we need to use `find_python3` in this function to make sure this is defined, no?

alcaeus: The find-python3.sh script is invoked in L28 in this file, is there anything else that needs to be done? I'll add that this ran correctly after the change where before it aborted with the `jq: command not found` error

blink1073: That invocation only imports the function, I think we still need this block to be self-contained:

```bash
  echo "Finding Python3 binary..." 1>&2
  PYTHON="$(find_python3 2>/dev/null)"
  echo "Finding Python3 binary... done." 1>&2
```

alcaeus: Ah, I see. I missed that from the other file, thanks for pointing that out!

blink1073 (approved):
Thank you!

PR #438, opened 2024-05-03
DRIVERS-2904 Test against 8.0 builds

blink1073:


.evergreen/download-mongodb.sh
```diff
@@ -267,14 +277,12 @@ get_mongodb_download_url_for ()
              MONGODB_32="http://downloads.mongodb.org/linux/mongodb-linux-x86_64-rhel55${DEBUG}-${VERSION_32}.tgz"
              MONGODB_30="http://downloads.mongodb.org/linux/mongodb-linux-x86_64-rhel55${DEBUG}-${VERSION_30}.tgz"
       ;;
-      linux-sles-11*-x86_64)
```
alcaeus, line 270: Noted this moved to the bottom of the linux-sles list to be consistent with the ordering of other systems 👍

.evergreen/download-mongodb.sh
```diff
@@ -388,10 +420,18 @@ get_mongodb_download_url_for ()
              MONGODB_50="http://downloads.10gen.com/linux/mongodb-linux-x86_64-enterprise-debian11${DEBUG}-${VERSION_50}.tgz"
              # SERVER-62299 Added support for Debian 11 in 5.0.8 and 6.0.0-rc0
       ;;
+      linux-debian-12*)
```
alcaeus, line 423: Note that debian is not sorted in descending order but has too many versions for us to fix in this PR 😅

.evergreen/download-mongodb.sh
```diff
@@ -56,6 +56,7 @@ get_mongodb_download_url_for ()
    VERSION_MONGOSH="2.1.1"
    # Set VERSION_RAPID to the latest rapid release each quarter.
    VERSION_RAPID="7.3.1"
+   VERSION_80="v8.0-latest"
    VERSION_70="7.0.9"
    VERSION_60_LATEST="v6.0-latest"
```
alcaeus, line 61: Not relevant for this PR, but do you remember why we keep 6.0-latest around when we don't have 7.0-latest? Is there a test requirement for 6.0-latest?

alcaeus (approved):
LGTM, thank you!

.evergreen/download-mongodb.sh
blink1073: Looking at the history, I suspect we should have removed it when we added VERSION_60 stable.

PR #445, opened 2024-05-29
minor: fix ports in README.md

abr-egn:
Turns out the test servers are [spun up](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/csfle/start-servers.sh#L33) on 9000..9002, not 8000..8002 🙂


blink1073 (approved):
Thanks!

PR #448, opened 2024-06-07
Fix OIDC self-test

blink1073:
We use a different build backend now, so a direct invocation of setup.py doesn't work.


ShaneHarvey (approved):

ShaneHarvey:
> We use a different build backend now

Specifically this change: https://github.com/mongodb/mongo-python-driver/pull/1644

PR #449, opened 2024-06-07
Fix secrets_handling examples

jmikola:

PR #450, opened 2024-06-10
Use bash -c option to invoke command with its own args

jmikola:
The `-c` option is necessary to ensure the vault name arg is passed to `setup-secrets.sh` and not `bash` itself. This was overlooked in https://github.com/mongodb-labs/drivers-evergreen-tools/pull/449.


jmikola:
Applying this change to https://github.com/mongodb/mongo-php-driver/pull/1579 fixed [the build](https://spruce.mongodb.com/version/666734d7517b2b000786119b/tasks).

blink1073 (approved):
LGTM!

PR #456, opened 2024-07-17
Allow GitHub Action to run on Windows

alcaeus:
This PR makes some changes to the orchestration action and involved scripts to allow them to run on Windows.

To reduce duplication in the download script, I've combined Windows and cygwin through a fallback, however since PHP does its Windows testing in GitHub Actions I'd appreciate it if someone could double-check that the `case` statement is still valid. It doesn't throw a syntax error, so that's a start. When running on Windows in GitHub Actions, the OS identifier is `mingw64_nt-<version>-x86_64`. Since the path to the extraction tools doesn't match the path used on cygwin, I had to duplicate the config. However, since 7zip is in the PATH on Windows, we might be able to reduce duplication again if it's in the PATH on cygwin as well. Again, can't test this since I don't have any Evergreen tasks on Windows.

The action needed a couple of changes, and I also added an output variable for the `crypt_shared` path, which is necessary for us to run tests in. I used the cygwin logic to kill processes on Windows, but wasn't able to test this as the orchestration isn't running in a build. Here's a sample build in PHPC that used the new action version: https://github.com/alcaeus/mongo-php-driver/actions/runs/9958433378/job/27559930163


alcaeus:
> It doesn't throw a syntax error, so that's a start.

Scratch that, MacOS doesn't like the syntax. Looks like I'll be bringing back the duplication.

alcaeus:
> > It doesn't throw a syntax error, so that's a start.
> 
> Scratch that, MacOS doesn't like the syntax. Looks like I'll be bringing back the duplication.

Changing the shebang to bash fixed the issue, so I was able to get this done without duplication.

The PR is now ready for review, I added Windows to the test matrix here on GHA to protect against regressions.

blink1073 (approved):
Nice!

PR #459, opened 2024-07-23
Download correct crypt_shared version when starting serverless instance

alcaeus:
We currently download the rapid release for serverless, but this is prone to being outdated and needs manual updating. With this change, the serverless `create-instance` script attempts to download crypt_shared for the matching version. If this version is not available, the setup will fail, indicating there's a problem with the serverless version. Previously, you would sometimes get test failures (for example in CSFLE/QE tests) when versions don't match.

Tested with the PHP driver and confirmed that it downloads MongoDB 8.0 (our serverless instances are running 8.0.0-rc13).


blink1073 (approved):
LGTM, thanks!

PR #460, opened 2024-07-23
Use correct python binary in run-orchestration

alcaeus:
Surprisingly, this works on most operating systems, but Debian 12 doesn't like it.


blink1073:
"Debian 11 (bullseye) has removed the "python" package and the '/usr/bin/python' symlink due to the deprecation of Python 2. No packaged scripts should depend on the existence of '/usr/bin/python': if they do, that is a bug that should be reported to Debian. You can use the 'python-is-python3' or 'python-is-python2' packages to restore an appropriate '/usr/bin/python' symlink for third-party or legacy scripts"

blink1073 (approved):
LGTM!

PR #461, opened 2024-07-24
Run action tests on MacOS

alcaeus:
This adds testing on MacOS for the GitHub Action, ensuring that mongo-orchestration is able to start servers.


blink1073 (approved):
Thanks!

PR #470, opened 2024-08-12
DRIVERS-2956 Add Script to Cherry-Pick PRs

blink1073:
We won't have a way to test the new script in EVG until it is merged.  I verified that the script works when run manually locally.


Jibola (approved):
LGTM

PR #473, opened 2024-08-26
Usability improvements

blink1073:
- Allow PYTHON to be provided instead of using find_python3
- Better handling of binary directories - remove as needed
- Fix handling of `pathlib.Path().expanduser()` on Windows
- Add convenience Makefile targets for `run-server` and `stop-server`
- Add top level documentation for running servers.


.evergreen/atlas/atlas-utils.sh
```diff
@@ -64,9 +64,11 @@ check_deployment ()
   ATLAS_BASE_URL=${ATLAS_BASE_URL:-"https://account-dev.mongodb.com/api/atlas/v1.0"}
   TYPE=${DEPLOYMENT_TYPE:-"clusters"}
 
-  echo "Finding Python3 binary..." 1>&2
-  PYTHON="$(find_python3 2>/dev/null)"
-  echo "Finding Python3 binary... done." 1>&2
+  if [ -z "${PYTHON:-}" ]; then
```
NoahStapp, line 67: Could we make this into a function somehow? We duplicate this exact code block several times.

NoahStapp (changes requested):

.evergreen/atlas/atlas-utils.sh
blink1073: Done!

NoahStapp: Did you push a new change? I don't see it if so.

blink1073: Whoops, pushed.

NoahStapp (approved):

PR #486, opened 2024-09-04
Fix support for RHEL 7.0

blink1073:
Passing build: https://spruce.mongodb.com/version/66d8ad0bd0d8e80007870224/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/.evergreen/start-orchestration.sh
+++ b/.evergreen/start-orchestration.sh
@@ -35,8 +35,8 @@ venvcreate "${PYTHON:?}" venv
 echo "Creating virtual environment 'venv'... done."
 
 # Install from github to get the latest mongo-orchestration, fall back on published wheel.
-# The fallback was added to accommodate RHEL 7.0, whose version of glibc does not support the
-# hatchling backend used by mongo-orchestration.
+# The fallback was added to accommodate versions of Python 3 for which there is no compatible version
+# of the hatchling backend used by mongo-orchestration.
 python -m pip install -q --upgrade 'https://github.com/mongodb/mongo-orchestration/archive/master.tar.gz' || python -m pip install -q --upgrade mongo-orchestration
 python -m pip list
 cd $DRIVERS_TOOLS
--- a/.evergreen/start-orchestration.sh
+++ b/.evergreen/start-orchestration.sh
@@ -35,6 +35,8 @@ venvcreate "${PYTHON:?}" venv
 echo "Creating virtual environment 'venv'... done."
 
 # Install from github to get the latest mongo-orchestration, fall back on published wheel.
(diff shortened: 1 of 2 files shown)


.evergreen/start-orchestration.sh
```diff
@@ -34,8 +34,8 @@ echo "Creating virtual environment 'venv'..."
 venvcreate "${PYTHON:?}" venv
 echo "Creating virtual environment 'venv'... done."
 
-# Install from github to get the latest mongo-orchestration.
-python -m pip install -q --upgrade 'https://github.com/mongodb/mongo-orchestration/archive/master.tar.gz'
+# Install from github to get the latest mongo-orchestration, fall back on published wheel.
+python -m pip install -q --upgrade 'https://github.com/mongodb/mongo-orchestration/archive/master.tar.gz' || python -m pip install -q --upgrade mongo-orchestration
```
ShaneHarvey, line 38: We intentionally install from master, why fallback to the published one?

blink1073: On rhel 7.0 (but not rhel 7.9), we cannot install hatchling because of the version of glibc.

ShaneHarvey: Oh that's lame, could you add a comment explaining that?

ShaneHarvey: Wait why does hatchling install fail? It looks like a pure python library: https://pypi.org/project/hatchling/#files

blink1073: Ah, the real issue is that find_python3 is using Python 3.6, for which there is no matching version of hatchling.  I'll update the comment, because we could still end up in a whack-a-mole situation.

blink1073: Here's the failing build: https://spruce.mongodb.com/task/drivers_tools_tests_os_requires_50__os_requires_50~rhel70_small_auth~auth_ssl~ssl_test_5.0_standalone_patch_cf39786af0437b52a0f9829190d55b11bfbfe354_66d8ab2dfbe79f0007abf9fb_24_09_04_18_47_19/logs?execution=0

ShaneHarvey (approved):

.evergreen/start-orchestration.sh
ShaneHarvey: Ah so the real issue is that we're using Python 3.6 but mongo-orc@master requires Python 3.8+

ShaneHarvey:
We may have to do a new release of mongo-orc to get this to work. Otherwise the fallback won't work with newer mongod versions.

blink1073:
https://pypi.org/project/mongo-orchestration/0.9.0/

PR #488, opened 2024-09-06
PYTHON-4734 Make node download more robust

blink1073:


nbbeeken:
Hey @blink1073, this caught my eye, of course, the Node team suffers from the same intermittent failure. The folks on mongosh have a more powerful retry approach: 

https://github.com/mongodb-js/mongosh/blob/57a904a1acb7821eee4e4cde98ff779b87c612b1/.evergreen/retry-with-backoff.sh#L13

Any interest in porting that here? Then the Node team can migrate to this copy of the script. 🙂

blink1073:
Ooh nice, yeah, let's use that.

NoahStapp (approved):

nbbeeken (approved):

ShaneHarvey:
Can't we use curl's built in retry flag?
```
--retry <num>

If a transient error is returned when curl tries to perform a transfer, it retries this number of times before giving up. Setting the number to 0 makes curl do no retries (which is the default). Transient error means either: a timeout, an FTP 4xx response code or an HTTP 408, 429, 500, 502, 503 or 504 response code.

When curl is about to retry a transfer, it first waits one second and then for all forthcoming retries it doubles the waiting time until it reaches 10 minutes which then remains delay between the rest of the retries.
```
https://curl.se/docs/manpage.html

blink1073:
We were already using the retry flag on the node download, it was still erroring out for other reasons.

ShaneHarvey:
What reasons? perhaps we want to use `--retry` plus `--retry-all-errors`?

ShaneHarvey:
Oh `--retry-all-errors` requires curl 7.71.0 released Jun 2020 which might be too new and not portable on our hosts.

blink1073:
That, and we're using this for other things that need to be retried.

ShaneHarvey:
The problem is that the new retry_with_backoff  function breaks when `set -e` is enabled.

blink1073:
Let's revert for now

PR #493, opened 2024-09-13
fix: make function syntax compatible with sh

W-A-James:
Change function declaration syntax used to be compatible with sh.


blink1073 (approved):
Thank you!

PR #497, opened 2024-09-17
RUST-2035 Fix invocation of mongosh for requireApiVersion

abr-egn:
RUST-2035

Example [failing run](https://parsley.mongodb.com/evergreen/mongo_rust_driver_stable_api_test_latest_standalone_61621db9719653b888652f1a56387009683fe280_24_09_13_13_56_55/0/task?bookmarks=0,1174&shareLine=1001) and [passing run](https://spruce.mongodb.com/version/66e9b4e75ae3be000756f058/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) against my fork with the fix.

I'm honestly not sure how this worked before; `mongosh` is [moved into](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/download-mongodb.sh#L751) the `$MONGODB_BINARIES` path by the install script and isn't in `$PATH`.

### DIFF
--- a/.evergreen/run-orchestration.sh
+++ b/.evergreen/run-orchestration.sh
@@ -173,5 +173,5 @@ EOT
 
 # Set the requireApiVersion parameter
 if [ ! -z "$REQUIRE_API_VERSION" ]; then
-  mongosh $URI $MONGO_ORCHESTRATION_HOME/require-api-version.js
+  $MONGODB_BINARIES/mongosh $URI $MONGO_ORCHESTRATION_HOME/require-api-version.js
 fi


blink1073:
What if we instead add to to PATH here? https://github.com/mongodb-labs/drivers-evergreen-tools/blob/4ff62cd75cdebe4ce166fe0003e364eaa310fb0e/.evergreen/handle-paths.sh#L56

blink1073:
I can follow up with a PR with a REQUIRE_API_VERSION test in drivers-tools as well.  I'll also need to add a test for invoking `download-mongodb.sh` directly to cover @eramongodb's case as well.

abr-egn:
> What if we instead add to to PATH here?
> 
> https://github.com/mongodb-labs/drivers-evergreen-tools/blob/4ff62cd75cdebe4ce166fe0003e364eaa310fb0e/.evergreen/handle-paths.sh#L56

That would work as well, but I'm generally more a fan of keeping state explicit 🙂  It's also more consistent with how mongosh is called from [other scripts](https://github.com/search?q=repo%3Amongodb-labs%2Fdrivers-evergreen-tools%20%22%24MONGODB_BINARIES%2Fmongosh%22&type=code) in this repo.

blink1073 (approved):
Thanks!

blink1073:
[test-oidc-k8s-local](https://spruce.mongodb.com/task/drivers_tools_tests_oidc_test_oidc_k8s_local_patch_4ff62cd75cdebe4ce166fe0003e364eaa310fb0e_66e9ba1f3406bf0007da2785_24_09_17_17_19_28?execution=0) is a known flakey test.

PR #498, opened 2024-09-17
Fix path handling in download-mongodb

blink1073:
I tested this by cloning the c driver and invoking the integration script.  The way that script gets invoked is pretty gnarly and unexpected.  The path ends up being `.evergreen/scripts/../../drivers-evergreen-tools/.evergreen/download-mongodb.sh`, and I'm not sure how to get that into a suitable path.  I'll have to follow up with a more robust solution.

### DIFF
--- a/.evergreen/download-mongodb.sh
+++ b/.evergreen/download-mongodb.sh
@@ -827,6 +827,10 @@ download_and_extract_crypt_shared ()
    mkdir crypt_shared_download
    cd crypt_shared_download
 
+   echo "HI HI $(pwd)"
+   # shellcheck disable=SC3028
+   echo "HELLO $0"
+   echo "HELLO $(realpath $0)"
    # shellcheck disable=SC3028
    SCRIPT_DIR=$(dirname ${BASH_SOURCE:-$0})
    "$SCRIPT_DIR/retry-with-backoff.sh" curl $MONGO_CRYPT_SHARED_DOWNLOAD_URL --output crypt_shared-binaries.tgz
--- a/.evergreen/download-mongodb.sh
+++ b/.evergreen/download-mongodb.sh
@@ -695,7 +695,6 @@ curl_retry ()
     { curl --fail -sS --max-time 300 "$@" && return 0; } || sleep $i
   done
   return 1
-  done
 }
 
 # download_and_extract_package downloads a MongoDB server package.
--- a/.evergreen/download-mongodb.sh
+++ b/.evergreen/download-mongodb.sh
@@ -697,7 +697,6 @@ download_and_extract_package ()
 
    # shellcheck disable=SC3028
    SCRIPT_DIR=$(dirname ${BASH_SOURCE:-$0})
-   . $SCRIPT_DIR/handle-paths.sh
 
    if [ -n "${MONGODB_BINARIES:-}" ]; then
       cd "$(dirname "$(dirname "${MONGODB_BINARIES:?}")")"
@@ -729,7 +728,6 @@ download_and_extract_mongosh ()
 
    # shellcheck disable=SC3028
    SCRIPT_DIR=$(dirname ${BASH_SOURCE:-$0})
-   . $SCRIPT_DIR/handle-paths.sh
 
    if [ -z "$MONGOSH_DOWNLOAD_URL" ]; then
       get_mongodb_download_url_for "$(get_distro)" latest false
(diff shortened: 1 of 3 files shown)


blink1073:
Worst case I'll follow up with an inline retry function because we're already seeing a failure with:

````
curl: (92) HTTP/2 stream 1 was not closed cleanly: INTERNAL_ERROR (err 2)
Error: Process completed with exit code 92.
```

blink1073:
test-oidc-k8s-local failure is expected, I had to tear down the cluster today

.evergreen/download-mongodb.sh
```diff
@@ -695,18 +695,14 @@ download_and_extract_package ()
...
-
    if [ -n "${MONGODB_BINARIES:-}" ]; then
       cd "$(dirname "$(dirname "${MONGODB_BINARIES:?}")")"
    else
       cd $DRIVERS_TOOLS
    fi
 
    echo "Installing server binaries..."
-   "$SCRIPT_DIR/retry-with-backoff.sh" curl $MONGODB_DOWNLOAD_URL --output mongodb-binaries.tgz
+   curl $MONGODB_DOWNLOAD_URL --output mongodb-binaries.tgz
```
ShaneHarvey, line 705: We should be using curl_retry here. curl_retry is defined in this file.

ShaneHarvey (changes requested):

eramongodb (commented):
> retry-with-backoff.sh

Has curl's native command line flags been considered?

> `--retry <num>`
>
> If a transient error is returned when curl tries to perform a transfer, it will retry this number of times before giving up. Setting the number to 0 makes curl do no retries (which is the default). Transient error means either: a timeout, an FTP 4xx response code or an HTTP 408, 429, 500, 502, 503 or 504 response code.
>
> **_When curl is about to retry a transfer, it will first wait one second and then for all forthcoming retries it will double the waiting time until it reaches 10 minutes which then will be the delay between the rest of the retries._**

This behavior seems completely equivalent to the emulated behavior of the `retry-with-backoff.sh` script. See also `--retry-connrefused` and `--retry-all-errors`.

.evergreen/download-mongodb.sh
ShaneHarvey: Oh you removed curl_retry in PYTHON-4734. How about we just revert that commit entirely (again)?

ShaneHarvey:
We previously used curl's built in retries but not all curls we use are new enough because we test on so many different/old hosts. See PYTHON-3721 and DRIVERS-2625.

blink1073:
Updated to restore `curl_retry`.  I'll make sure to test the c driver if changing this file in the future.

ShaneHarvey (approved):
LGTM assuming the tests pass.

eramongodb (approved):
Minor feedback regarding `curl_retry`. Otherwise, as far as the C Driver is concerned, latest changes are LGTM.

.evergreen/download-mongodb.sh
```diff
@@ -688,25 +688,27 @@ set_url_win32 ()
   MONGODB_24="https://fastdl.mongodb.org/win32/mongodb-win32-i386${DEBUG}-${VERSION_24}.zip"
 }
 
+# curl_retry runs curl with up to three retries, retrying any error.
+curl_retry ()
+{
+  for i in 2 4 8; do curl --fail -sS --max-time 300 "$@" && break || sleep $i;
```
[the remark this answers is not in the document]
blink1073: I'm afk for the next hour.  feel free to merge if you're blocked

.evergreen/download-mongodb.sh
```diff
@@ -688,25 +688,30 @@ set_url_win32 ()
...
 }
 
+# curl_retry runs curl with up to three retries, retrying any error.
+curl_retry ()
+{
+  for i in 1 2 4 8 16; do
+    { curl --fail -sS --max-time 300 "$@" && return 0; } || sleep $i
+  done
+  return 1
+  done
```
eramongodb, line 698:
```suggestion
```

.evergreen/download-mongodb.sh
```diff
@@ -688,25 +688,29 @@ set_url_win32 ()
   MONGODB_24="https://fastdl.mongodb.org/win32/mongodb-win32-i386${DEBUG}-${VERSION_24}.zip"
 }
 
+# curl_retry runs curl with up to three retries, retrying any error.
```
eramongodb, line 691:
```suggestion
# curl_retry emulates running curl with `--retry 5` and `--retry-all-errors`.
```

PR #500, opened 2024-09-20
Combine fetch source and prepare resources in EVG config.yml

Jibola:


blink1073:
This is hard to review because of the yaml reformatter.  Can we do that as a separate PR so we can add it to git-blame-ignore-revs?

Jibola:
> This is hard to review because of the yaml reformatter. Can we do that as a separate PR so we can add it to git-blame-ignore-revs?

Whoops. Thanks for pointing that out. Didn't mean to place in the autoformatting.

blink1073:
> Please set the PROJECT_DIRECTORY environment variable.

blink1073:
There is a merge conflict

blink1073:
It looks like UPLOAD_BUCKET isn't being set properly, maybe add some debug logging?

Jibola:
need to remove additional calls to `prepare resources`

blink1073 (approved):
LGTM!

PR #509, opened 2024-09-30
Support handling multiple PIDs in killport

eramongodb:
Addresses the following possible error on certain distros (e.g. Ubuntu 18.04) due to `lsof` returning more than one PID:

```
start-orchestration.sh: line 76: kill: 4850
5401: arguments must be process or job IDs
```

Extends all find-and-kill-PID routines to support the possibility of multiple PIDs being returned by the PID query command.


blink1073 (approved):
LGTM!

PR #512, opened 2024-10-01
killport: wait up to 1 min for server shutdown before kill-and-continue

eramongodb:


.evergreen/start-orchestration.sh
```diff
@@ -67,17 +67,21 @@ killport() {
 
   if [[ "${OSTYPE:?}" == cygwin || "${OSTYPE:?}" == msys ]]; then
     for pid in $(netstat -ano | grep ":$port .* LISTENING" | awk '{print $5}' | tr -d '[:space:]'); do
-      taskkill /F /T /PID "$pid" || true
```
ShaneHarvey, line 70: `/F` already force kills so we shouldn't need to wait here.

.evergreen/start-orchestration.sh
```diff
@@ -67,17 +67,21 @@ killport() {
...
+      kill "$pid" || true
+      timeout 60s bash -c "while kill -0 \"$pid\" 2>/dev/null; do sleep 1; done" || kill -9 "$pid" || true
     done
   elif [ -x "$(command -v lsof)" ]; then
     for pid in $(lsof -t "-i:$port" || true); do
       kill "$pid" || true
+      timeout 60s bash -c "while kill -0 \"$pid\" 2>/dev/null; do sleep 1; done" || kill -9 "$pid" || true
```
ShaneHarvey, line 76: The reason I changed it to SIGTERM was to give MO a chance to cleanup its mongo processes but that doesn't seem to work. If we're going back to `kill -9` (SIGKILL) instead of SIGTERM then can we remove the timeout logic?

.evergreen/start-orchestration.sh
eramongodb: This PR proposes using `kill` instead of `taskkill` to extend graceful termination + timeout->kill to Windows distros as well. Please re-review.

.evergreen/start-orchestration.sh
eramongodb: The use of `-9` in the while loop was a typo. `-0` was meant to be used instead. Please re-review.

eramongodb: Quoting @nbbeeken from Slack discussions:

> The switch to sigterm was intentional because mongo-orch will try to clean up the servers gracefully, but maybe there's a bug in that logic and we should send -9 anyway :blob-sad:  you try to be a 'good citizen' and look what it gets you

If we want to revert to the simpler "just SIGKILL" routines we had before, that would also address the flaky "address already in use" problem this PR is trying to solve.

blink1073: In Jupyter we use the pattern of SIGTERM, wait on pid, SIGKILL as well.  I think it is a good pattern in general.

ShaneHarvey: Yeah I'd prefer SIGKILL at this point because it should save time (don't need to wait 60 seconds) and will make this code easier to read.

.evergreen/start-orchestration.sh
blink1073: Should we revert this now as well?

eramongodb:
Updated PR so changes are to revert to using `SIGKILL` instead of `SIGTERM` + wait. The `taskkill -> kill` change is preserved as a simplification and consistency improvement.

.evergreen/start-orchestration.sh
eramongodb: I am in favor of keeping this change as a simplification + consistency improvement given this script is known to be Bash-executed. The less distro-specific patterns we need to depend on, the better.

blink1073 (approved):
LGTM!

blink1073 (approved):

.evergreen/start-orchestration.sh
ShaneHarvey: I'm concerned about this one. `taskkill /F /T ` kills the process and all child processes which seems safer than just `kill -SIGKILL`. I don't want to make windows less stable after this change so I'd prefer we keep it using taskkill

eramongodb: Done.

ShaneHarvey (approved):

PR #520, opened 2024-10-07
DRIVERS-2999 Use isolated orchestration file

blink1073:
This prevents us from modifying the orchestration files in place that use ABSOLUTE_PATH_REPLACEMENT_TOKEN.

Tested with the C driver: https://spruce.mongodb.com/version/67041e9b7f1bfe000734ae39/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/.evergreen/run-orchestration.sh
+++ b/.evergreen/run-orchestration.sh
@@ -53,7 +53,7 @@ printf '%s' $MONGODB_BINARIES | $PYTHON -c 'import json,sys; print(json.dumps({"
 
 # Copy client certificate because symlinks do not work on Windows.
 mkdir -p ${MONGO_ORCHESTRATION_HOME}/lib
-cp ${DRIVERS_TOOLS}/.evergreen/x509gen/client.pem ${MONGO_ORCHESTRATION_HOME}/lib/client.pem || true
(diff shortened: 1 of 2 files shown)


ShaneHarvey (approved):
LGTM but could you create a DRIVERS ticket to track this?

.evergreen/run-orchestration.sh
```diff
@@ -119,13 +119,15 @@ else
 fi
 echo "ORCHESTRATION_FILE=$ORCHESTRATION_FILE"
 
+# Copy the orchestration file so we can override it.
+cp $ORCHESTRATION_FILE $MONGO_ORCHESTRATION_HOME/config.json
```
eramongodb, line 123:
```suggestion
cp -f "$ORCHESTRATION_FILE" "$MONGO_ORCHESTRATION_HOME/config.json"
```
Should this command have an `-f` flag (overwrite existing `config.json`)?

eramongodb (approved):
Minor feedback; otherwise, LGTM.

FYI https://github.com/mongodb/mongo-c-driver/pull/1742 refactored the C Driver EVG config to use `run-orchestration.sh` instead of its own custom logic (including use of `ABSOLUTE_PATH_REPLACEMENT_TOKEN`, which is no longer present) so it should be less prone to breakage from DET script refactors from now on.

PR #530, opened 2024-10-25
DRIVERS-3002 Update mongodl validation

blink1073:
Follow up to #529

### DIFF
--- a/.evergreen/mongodl.py
+++ b/.evergreen/mongodl.py
@@ -410,7 +410,7 @@ class CacheDB:
                 for distro in DISTRO_ID_TO_TARGET.values():
                     if target in list(distro.values()):
                         found = True
-                if not found and target not in ['macos', 'windows']:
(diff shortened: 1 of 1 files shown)


.evergreen/mongodl.py
```diff
@@ -410,7 +410,7 @@ def _import_json_data(self, data: 'Any') -> None:
                 for distro in DISTRO_ID_TO_TARGET.values():
                     if target in list(distro.values()):
                         found = True
-                if not found and target not in ['macos', 'windows']:
+                if not found and target not in ['linux_i686', 'linux_x86_64', 'osx', 'macos', 'windows']:
```
eramongodb, line 413: Recommend adding a comment and/or extracting this list into a named constant variable describing when/why the listed entries should be ignored despite being missing from the `DISTRO_ID_TO_TARGET` map.

eramongodb (approved):

.evergreen/mongodl.py
blink1073: Done!

eramongodb (approved):

PR #533, opened 2024-10-28
DRIVERS-3019 Add ruff linter and apply fixes

blink1073:
In follow-up PRs, we will add ruff format and then add a [.git-blame-ignore-revs](https://github.com/mongodb/mongo-python-driver/blob/master/.git-blame-ignore-revs) file to ignore the formatting changes in git blame.


aclark4life:
Nice lint, LGTM thanks!

aclark4life (approved):

PR #535, opened 2024-10-28
DRIVERS-2415 Use a certfile for Azure OIDC login

blink1073:
https://learn.microsoft.com/en-us/cli/azure/azure-cli-sp-tutorial-7?tabs=bash#reset-credentials-returning-output-to-the-console-and-log-file

After being prompted by Azure:

```
ERROR: AADSTS7000222: The provided client secret keys for app 'xxx' are expired. Visit the Azure portal to create new keys for your app: https://aka.ms/NewClientSecret, or consider using certificate credentials for added security: https://aka.ms/certCreds.
```

### DIFF
--- a/.evergreen/auth_oidc/azure/create-and-setup-vm.sh
+++ b/.evergreen/auth_oidc/azure/create-and-setup-vm.sh
@@ -17,6 +17,7 @@ pushd $SCRIPT_DIR
 # Set defaults.
 export AZUREKMS_PUBLICKEYPATH="$SCRIPT_DIR/keyfile.pub"
 export AZUREKMS_PRIVATEKEYPATH="$SCRIPT_DIR/keyfile"
+export AZUREKMS_CERTFILE="$SCRIPT_DIR/cert.pem"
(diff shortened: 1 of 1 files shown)


.evergreen/auth_oidc/azure/create-and-setup-vm.sh
```diff
@@ -27,8 +28,12 @@ if [ ! -f ./secrets-export.sh ]; then
 fi
 source ./secrets-export.sh
 
+echo "${AZUREOIDC_CERT}" | base64 --decode > $AZUREKMS_CERTFILE
```
kevinAlbs, line 31: Suggest adding check for `AZUREOIDC_CERT` in the "Ensure required variables are set" block above.

kevinAlbs (approved):
LGTM with suggested check.

.evergreen/auth_oidc/azure/create-and-setup-vm.sh
blink1073: The cert is found from the vault, the only required manual variable for this test is AZUREOIDC_VMNAME_PREFIX.

PR #543, opened 2024-11-04
add missing return to mock KMS server

kevinAlbs:
Intended to address observed errors in C# tests: "KMS response parser error with status 200, error: 'Unexpected: exceeded content length" seemingly caused by multiple HTTP responses returned on the same socket side-by-side.


BorisDog (approved):
LGTM + 
A question regarding `return` in `Unknown AWS Operation` case for your consideration.

PR #544, opened 2024-11-06
DRIVERS-2949 Test server for happy eyeballs

abr-egn:
DRIVERS-2949

This adds a relatively simple server that drivers can use to test Happy Eyeballs connection behavior, as well as a sample client to exercise the server behavior.  Use of said server can be seen in [this repo](https://github.com/abr-egn/mongo-rust-driver/tree/DRIVER-2949/evergreen-tools-server), particularly [this chunk of Evergreen config](https://github.com/abr-egn/mongo-rust-driver/blob/DRIVER-2949/evergreen-tools-server/.evergreen/config.yml#L1732) and [this test](https://github.com/abr-egn/mongo-rust-driver/blob/DRIVER-2949/evergreen-tools-server/src/test/happy_eyeballs.rs).


blink1073:
Can you please include a README with that example yaml?

abr-egn:
> Can you please include a README with that example yaml?

Done; also shifted the rest of the usage instructions from the top of `server.py` into the README since it made sense to me to keep them all together.

.evergreen/happy_eyeballs/README.md
```diff
@@ -0,0 +1,57 @@
+# Happy Eyeballs test scripts
+
+This folder contains a test server ([`server.py`](server.py)) for driver Happy Eyeballs TCP connection behavior.  It also has has [`client.py`](client.py), a simple client for that server that can be useful for debugging but won't be needed in most cases.
+
+**NOTE**: This server relies on network stack behavior present in Windows and MacOS but not Linux, so any tests using it must only be run on those two OSen.
```
blink1073, line 5:
```suggestion
**NOTE**: This server relies on network stack behavior present in Windows and MacOS but not Linux, so any tests using it must only be run on those two OSes.
```

blink1073 (approved):

kevinAlbs (commented):
Consider having IPv4 and IPv6 listen on the same port and using OP_MSG for messages. See [this branch](https://github.com/mongodb-labs/drivers-evergreen-tools/compare/master...kevinAlbs:drivers-evergreen-tools:with-OPMSG.happy-eyeballs-test?expand=1) for a proposal.

It adds complexity to the test server, but enables testing with public driver API:

```python
import socket
from pymongo import MongoClient
from pymongo.server_api import ServerApi

control = socket.create_connection(("localhost", 10036))
control.sendall(b"\x04") # Request slow IPv4 responses.
assert control.recv(1) == b"\x01" # Receive OK.
port = int.from_bytes(control.recv(2), 'big') # Receive port.

client = MongoClient(f"localhost:{port}", server_api=ServerApi('1')) # Use API version to force OP_MSG.
got = client["db"].command({"ping": 1})
print(got) # Prints: {'ok': 1, 'ip': 'ipv4'}
client["db"].command({"shutdown": 1}) # Shutdown server.
```

This may ease testing for drivers. The current C implementation of the single-threaded client (which attempts to implement Happy Eyeballs) does not-yet have a way to [mock DNS results](https://github.com/mongodb/mongo-c-driver/blob/3b0078e8d9150f193da2cb47f6904057cfcf3854/src/libmongoc/src/mongoc/mongoc-topology-scanner.c#L883).

abr-egn:
> Consider having IPv4 and IPv6 listen on the same port and using OP_MSG for messages. See [this branch](https://github.com/mongodb-labs/drivers-evergreen-tools/compare/master...kevinAlbs:drivers-evergreen-tools:with-OPMSG.happy-eyeballs-test?expand=1) for a proposal.

I don't think this is a good direction to go.  Your proposed branch nicely minimizes the added complexity in the test server, but this is inherently a timing-based test, and adding the overhead of client construction and message construction and parsing introduces a lot of room for both variance between drivers and test flakiness.

> It adds complexity to the test server, but enables testing with public driver API:
> 
> ```python
> import socket
> from pymongo import MongoClient
> from pymongo.server_api import ServerApi
> 
> control = socket.create_connection(("localhost", 10036))
> control.sendall(b"\x04") # Request slow IPv4 responses.
> assert control.recv(1) == b"\x01" # Receive OK.
> port = int.from_bytes(control.recv(2), 'big') # Receive port.
> 
> client = MongoClient(f"localhost:{port}", server_api=ServerApi('1')) # Use API version to force OP_MSG.
> got = client["db"].command({"ping": 1})
> print(got) # Prints: {'ok': 1, 'ip': 'ipv4'}
> client["db"].command({"shutdown": 1}) # Shutdown server.
> ```

I think this actually illustrates why I don't want to go with this approach - it's hard to tell if this is actually testing Happy Eyeballs or not.  Presumably the `MongoClient` construction is doing name resolution on `localhost`, but it's very possible that's just returning `127.0.0.1` and not `::1`.  If it were properly racing the two connections, the response gotten should have been `{'ok': 1, 'ip': 'ipv6'}`, which looks like isn't what happened.

> This may ease testing for drivers. The current C implementation of the single-threaded client (which attempts to implement Happy Eyeballs) does not-yet have a way to [mock DNS results](https://github.com/mongodb/mongo-c-driver/blob/3b0078e8d9150f193da2cb47f6904057cfcf3854/src/libmongoc/src/mongoc/mongoc-topology-scanner.c#L883).

From discussion, it looks like the C driver is unique in this being a stumbling block.

kevinAlbs (approved):
LGTM

> If it were properly racing the two connections, the response gotten should have been `{'ok': 1, 'ip': 'ipv6'}`, which looks like isn't what happened.

I expect pymongo is not doing Happy Eyeballs. An analogous test with C's Happy Eyeballs implementation gets back the expected "ipv6".

> From discussion, it looks like the C driver is unique in this being a stumbling block.

Agreed, since this seems unique to C, I prefer the server implementation. Linking to [slack thread](https://mongodb.slack.com/archives/C72LB5RPV/p1731342948905369).

ShaneHarvey:
That's correct, Pymongo does not yet implement happy eyeballs. We connect sequentially in whatever order getaddrinfo returns (except for 'localhost' which always uses ipv4).

PR #551, opened 2024-11-15
test: increase gke pod memory

durran:
Node testing on GKE runs out of memory with the 2GB limit. Increased to 4GB.


blink1073 (approved):
LGTM!

PR #552, opened 2024-11-15
DRIVERS-3032 Add standalone clis for top level python scripts

blink1073:
cc @baileympearson 

This handles just the top level python scripts.  Using this pattern we can add clis to the the folders that currently use the `activate<foo>.sh` scripts, and a python cli for run-orchestration.


baileympearson (approved):
I tested this locally and confirmed that it works.  I added the install directory to my path and can run drivers-evergreen-tools Python scripts directly

NoahStapp (approved):

PR #570, opened 2025-01-06
Properly daemonize the csfle servers

blink1073:
I tested this with the Go driver, which had use the background approach.  This will allow us to move the encryption setup to a task in the Python driver, without having to use the background approach.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -520,6 +520,18 @@ functions:
       params:
         binary: bash
         include_expansions_in_env: [AWS_SECRET_ACCESS_KEY, AWS_ACCESS_KEY_ID, AWS_SESSION_TOKEN]
+        args: [src/.evergreen/csfle/setup.sh]
+    - command: subprocess.exec
+      type: test
+      params:
+        binary: bash
+        include_expansions_in_env: [AWS_SECRET_ACCESS_KEY, AWS_ACCESS_KEY_ID, AWS_SESSION_TOKEN]
+        args: [src/.evergreen/csfle/teardown.sh]
+    - command: subprocess.exec
+      type: test
+      params:
+        binary: bash
+        include_expansions_in_env: [AWS_SECRET_ACCESS_KEY, AWS_ACCESS_KEY_ID, AWS_SESSION_TOKEN]
         args: [src/.evergreen/tests/test-csfle.sh]
 
   "run cli test partial":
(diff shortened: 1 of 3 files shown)


.evergreen/config.yml
```diff
@@ -515,6 +515,18 @@ functions:
...
+      params:
+        binary: bash
+        include_expansions_in_env: [AWS_SECRET_ACCESS_KEY, AWS_ACCESS_KEY_ID, AWS_SESSION_TOKEN]
+        args: [src/.evergreen/csfle/setup.sh]
+    - command: subprocess.exec
+      type: test
+      params:
+        binary: bash
+        include_expansions_in_env: [AWS_SECRET_ACCESS_KEY, AWS_ACCESS_KEY_ID, AWS_SESSION_TOKEN]
+        args: [src/.evergreen/csfle/teardown.sh]
```
ShaneHarvey, line 529: Does this work? It looks like we are calling setup.sh then immediately calling teardown.sh. Shouldn't the teardown happen later after running the test suite?

.evergreen/csfle/start-servers.sh
```diff
@@ -30,50 +30,60 @@ done
 . ./activate-kmstlsvenv.sh
 
 # The -u options forces the stdout and stderr streams to be unbuffered.
-# TMPDIR is required to avoid "AF_UNIX path too long" errors.
+COMMAND="python -u"
+if [ "$(uname -s)" != "Darwin" ]; then
+  # On linux and windows host, we need to use nohup to daemonize the process
+  # and prevent the task from hanging.
+  # The macos hosts do not support nohup.
```
ShaneHarvey, line 37: What happens on macos?

ShaneHarvey (changes requested):

.evergreen/config.yml
blink1073: This is just to ensure that setup doesn't hang.  The test suite itself does its own setup and teardown for each python version.

.evergreen/csfle/start-servers.sh
blink1073: It does not hang on macos, as confirmed by the test run: https://spruce.mongodb.com/task/drivers_tools_tests_all__os_fully_featured~macos_14_auth~auth_ssl~nossl_test_csfle_patch_eadf5c21e411d5582e52b43cfe9e2ffd26274fb8_677beac5430c5d0007e4745d_25_01_06_14_37_57/logs?execution=0&sortBy=STATUS&sortDir=ASC.

.evergreen/config.yml
ShaneHarvey: Can we add a comment explaining this and also the purpose of test-csfle.sh below?

ShaneHarvey (approved):

PR #575, opened 2025-01-09
Add section concerning amd64 with rancher desktop

prestonvasquez:
Rancher desktop's `27.4.1-rd` does not do a great job of selecting the best architecture for an image, rather defaulting to the machine's architecture. This causes strange behavior, especially for dependencies that are specific to amd64 (libsnmp35, for example). Add section noting this to the docker README.md.


blink1073 (approved):
Thanks!

PR #585, opened 2025-01-29
Fix support for MongoDB 3.6

blink1073:
The PyMongo 4.11 release dropped support for MongoDB 3.6, so we need to install an older version of PyMongo when testing against 3.6.

### DIFF
--- a/.evergreen/install-cli.sh
+++ b/.evergreen/install-cli.sh
@@ -48,23 +48,24 @@ fi
 pushd $1 > /dev/null
 
 # Add support for MongoDB 3.6, which was dropped in pymongo 4.11.
-EXTRA_ARGS=()
+EXTRA_ARGS=""
 if [ "${MONGODB_VERSION:-latest}" == "3.6" ]; then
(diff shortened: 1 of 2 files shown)


.evergreen/install-cli.sh
```diff
@@ -47,18 +47,25 @@ fi
 
 pushd $1 > /dev/null
 
+# Add support for MongoDB 3.6, which was dropped in pymongo 4.11.
+EXTRA_ARGS=""
+if [ "${MONGODB_VERSION:-latest}" == "3.6" ]; then
+  EXTRA_ARGS_ARR=(--with "pymongo<4.11")
+  EXTRA_ARGS="${EXTRA_ARGS_ARR[*]}"
```
Jibola, line 54: NIT: could this just be one line?

Jibola (approved):

.evergreen/install-cli.sh
blink1073: No, shellcheck didn't like it

PR #588, opened 2025-01-31
Added sensible defaults in drivers_orchestration when env variables used

caseyclements:
Adds sensible defaults for version (latest) and topology (server) when en passed as env vars not cli args.
This was causing failures in csharp driver.


sanych-sun (approved):
LGTM

PR #589, opened 2025-02-06
DRIVERS-3032 Add support for ubuntu2204-arm64

blink1073:


.evergreen/install-cli.sh
```diff
@@ -65,7 +65,7 @@ if [ "Windows_NT" == "${OS:-}" ]; then
   done
   rm -rf $TMP_DIR
 else
-  UV_TOOL_BIN_DIR=$(pwd) uv tool install -q ${EXTRA_ARGS} --python "$(which python)" --force --editable .
+  UV_TOOL_BIN_DIR=$(pwd) python -m uv tool install -q ${EXTRA_ARGS} --python "$(which python)" --force --editable .
```
eramongodb, line 68: Does the call to `uv` on line 61 not also need a `python -m` prefix (consistency)?

eramongodb (approved):

.evergreen/install-cli.sh
blink1073: Done

vector-of-bool: I suspect that this change may have broken some things: https://spruce.mongodb.com/task/mongo_c_driver_smoke_link_with_cmake_ssl_patch_a91d6f6ad6e650a1dee9e99c3c2dcbafd39e060a_67a65e5d071d5a0007bedac5_25_02_07_19_26_22/logs?execution=2

Am not certain, though.

PR #592, opened 2025-02-07
DRIVERS-3032 ensure uv is installed after venv activation

eramongodb:
Followup to https://github.com/mongodb-labs/drivers-evergreen-tools/pull/591 and https://github.com/mongodb-labs/drivers-evergreen-tools/pull/591.

Several issues appear to be related to the errors that motivated prior PRs:

- If `venv` already exists, it may be activated without a call to `pip install uv` even if `uv` has not yet been installed into that venv.
- uv may not necessarily be installed directly into `venv/bin` or `venv/Script`. It is possible for the binary to reside in a deeper subdirectory such as "site-packages", which in turn may also be within a Python-versioned subdirectory, such as "python-3.12" (given uv is designed to support multiple Python versions). The `pushd $1` after activating the virtual environment may additionally break `$PATH` expectations (although I would expect/assume calling `pip` or `python -m pip` within an activate environment should be resiliant to this...).

Regardless, it seems sufficent to 1) ensure `uv` is checked and installed after activation (regardless whether the environment already existed or not, and 2) store and use the full path to the resulting binary (via `which`) to account for changing working directories.

### DIFF
--- a/.evergreen/install-cli.sh
+++ b/.evergreen/install-cli.sh
@@ -40,10 +40,19 @@ if [ ! -d $SCRIPT_DIR/venv ]; then
   echo "Creating virtual environment 'venv'..."
   venvcreate "${PYTHON:?}" venv
   echo "Creating virtual environment 'venv'... done."
-  python -m pip install -q uv
 else
   venvactivate venv
 fi
+
+if ! command -v uv >/dev/null; then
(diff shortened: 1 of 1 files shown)


blink1073:
Passing with python driver, merging, thanks!

blink1073 (approved):

PR #597, opened 2025-02-11
fix: node-gyp requires 3.12+ python version

nbbeeken:
This won't actually guarantee 3.12 plus but it will get us a lot closer.


nbbeeken:
https://spruce.mongodb.com/version/67aba16f3dc83500075d42bd/tasks

nbbeeken:
This does fix the python version issue, seems like we now find a python 3.13 which is perfect. The failure in the task is now back to the issue we're fixing in the linked driver PR where we need prebuild to pull in the N-API version parse fix.

durran (approved):

PR #618, opened 2025-03-04
Fix indentation in Evergreen yaml examples for k8s OIDC setup.

matthewdale:
Copy-pasting the current examples results in invalid Evergreen configs because some keys have the wrong indentation, creating child objects rather than sibling keys.


blink1073 (approved):
Thanks!

PR #626, opened 2025-03-10
PYTHON-5195 Allow OCSP server to be run as a daemon

blink1073:
Tested with https://github.com/mongodb/mongo-python-driver/pull/2190: https://spruce.mongodb.com/version/67cc94d4b26dff000739c152/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

and the C driver: https://spruce.mongodb.com/version/67ced2d64d80c40007a73649/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

Uses `waitress` to create a production flask server that can be run as a daemon on all platforms.
Adds tests for OCSP server on Mac, Windows, and RHEL.

### DIFF
--- a/.evergreen/ocsp/mock-ocsp-responder-requirements.txt
+++ b/.evergreen/ocsp/mock-ocsp-responder-requirements.txt
@@ -4,5 +4,6 @@ flask==2.2.5
 itsdangerous==2.1.2
 Jinja2==3.1.5
 MarkupSafe==2.1.4
-git+https://github.com/wbond/oscrypto.git@d5f3437
+oscrypto==1.3.0
+waitress==3.0.2
 Werkzeug==3.0.6
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -559,6 +559,13 @@ functions:
         binary: bash
         args: [src/.evergreen/tests/test-cli.sh]
 
+  "run ocsp test":
+    - command: subprocess.exec
+      type: test
+      params:
+        binary: bash
+        args: [src/.evergreen/tests/test-ocsp.sh]
+
   "teardown assets":
     - command: subprocess.exec
       params:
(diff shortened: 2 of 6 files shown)


NoahStapp (approved):

jmikola:
> Adds tests for OCSP server on Mac, Windows, and RHEL

@blink1073: I noted that this wasn't tested on Debian 12 hosts, but are you aware of any workaround for the following error when launching the OCSP responder?

```
oscrypto.errors.LibraryNotFoundError: Error detecting the version of libcrypto
```

See: [PHPC task failure](https://evergreen.mongodb.com/task_log_raw/mongo_php_driver_test_debian12_php_8.3_ocsp_test_1_ecdsa_delegate_8.0_patch_5160a8c82b00dce0093f01273f3de0b6589d76a5_67d1c94c21fc5400070507f2_25_03_12_17_50_24/0?type=T#L116) for more context.

I think this can be traced back to https://github.com/wbond/oscrypto/issues/78 and there appear to be some suggested workarounds that depend on a specific commit since it looks like the fix has yet to be tagged.

If that's not feasible, I can look into switching our OCSP tasks over to a different platform.

.evergreen/ocsp/mock-ocsp-responder-requirements.txt
```diff
@@ -4,5 +4,6 @@ flask==2.2.5
 itsdangerous==2.1.2
 Jinja2==3.1.5
 MarkupSafe==2.1.4
-git+https://github.com/wbond/oscrypto.git@d5f3437
```
jmikola, line 7: @blink1073: This removal is what appears to have caused the regression in PHPC. I noticed that this commit was mentioned in https://github.com/wbond/oscrypto/issues/78#issuecomment-1756601738 as a workaround for the upstream issue with oscrypto 1.3.0.

The original context for this change was https://github.com/blink1073/drivers-evergreen-tools/commit/4483bcca6e9084d124632d277ed45561f2f7464e, but it looks like we'd do well to include a comment here for future reference. Might be a good idea to reference the upstream issue.

blink1073: Hmm, I had removed the specific tag because I was seeing a warning from pip.  I'll revert to use the tag.

PR #631, opened 2025-03-12
PYTHON-5195 Use specific oscrypto tag version

blink1073:


blink1073:
Closing in favor of #632

PR #452, opened 2024-06-27
Use newer version of oscrypto for OCSP mock server

alcaeus:
The stable version of oscrypto has a [bug](https://github.com/wbond/oscrypto/issues/75) that prevents it from working with OpenSSL 3.0.10:

```
Traceback (most recent call last):
[2024/06/27 14:49:14.487]   File "/data/mci/91c84d5568444dead3a62ac8435331d4/drivers-tools/.evergreen/ocsp/ocsp_mock.py", line 13, in <module>
[2024/06/27 14:49:14.487]     import mock_ocsp_responder
[2024/06/27 14:49:14.487]   File "/data/mci/91c84d5568444dead3a62ac8435331d4/drivers-tools/.evergreen/ocsp/mock_ocsp_responder.py", line 56, in <module>
[2024/06/27 14:49:14.487]     from oscrypto import asymmetric
[2024/06/27 14:49:14.487]   File "/data/mci/91c84d5568444dead3a62ac8435331d4/drivers-tools/.evergreen/ocsp/ocspvenv/lib/python3.10/site-packages/oscrypto/asymmetric.py", line 19, in <module>
[2024/06/27 14:49:14.487]     from ._asymmetric import _unwrap_private_key_info
[2024/06/27 14:49:14.487]   File "/data/mci/91c84d5568444dead3a62ac8435331d4/drivers-tools/.evergreen/ocsp/ocspvenv/lib/python3.10/site-packages/oscrypto/_asymmetric.py", line 27, in <module>
[2024/06/27 14:49:14.487]     from .kdf import pbkdf1, pbkdf2, pkcs12_kdf
[2024/06/27 14:49:14.487]   File "/data/mci/91c84d5568444dead3a62ac8435331d4/drivers-tools/.evergreen/ocsp/ocspvenv/lib/python3.10/site-packages/oscrypto/kdf.py", line 9, in <module>
[2024/06/27 14:49:14.487]     from .util import rand_bytes
[2024/06/27 14:49:14.487]   File "/data/mci/91c84d5568444dead3a62ac8435331d4/drivers-tools/.evergreen/ocsp/ocspvenv/lib/python3.10/site-packages/oscrypto/util.py", line 14, in <module>
[2024/06/27 14:49:14.487]     from ._openssl.util import rand_bytes
[...]

PR #632, opened 2025-03-12
User newer oscrypto to fix detection of later OpenSSL versions

jmikola:
This was previously added in https://github.com/mongodb-labs/drivers-evergreen-tools/pull/452 but inadvertently reverted in https://github.com/mongodb-labs/drivers-evergreen-tools/pull/626


blink1073 (approved):
LGTM!

PR #634, opened 2025-03-13
PYTHON-5203 Use uv from Python toolchain if available

blink1073:
Tested with a full C driver patch build.  Failures appear to be unrelated.

https://spruce.mongodb.com/version/67d1ad2f8ad854000773caeb/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/.evergreen/find-python3.sh
+++ b/.evergreen/find-python3.sh
@@ -320,7 +320,7 @@ find_python3() (
 ensure_python3() {
   # Use "$DRIVERS_TOOLS_PYTHON".
   if command -v "${DRIVERS_TOOLS_PYTHON:-}"; then
-    echo "Using Python binary ${DRIVERS_TOOLS_PYTHON:?}" >&2
+    echo "Using Python binary ${DRIVERS_TOOLS_PYTHON:?}" 1>&2
     echo "${DRIVERS_TOOLS_PYTHON:?}"
     return
   fi
@@ -339,7 +339,7 @@ ensure_python3() {
     ;;
   esac
   if command -v "${python_binary:?}" >/dev/null; then
-    echo "Using Python binary ${python_binary:?}" >&2
+    echo "Using Python binary ${python_binary:?}" 1>&2
     echo "${python_binary:?}"
     return
   fi
--- a/.evergreen/find-python3.sh
+++ b/.evergreen/find-python3.sh
@@ -338,7 +338,7 @@ ensure_python3() {
     ;;
   esac
   if command -v "${python_binary:?}" >/dev/null; then
-    echo "${python_binary:?}"
+    echo "hello ${python_binary:?} hello"
     return
   fi
 
--- a/.evergreen/find-python3.sh
+++ b/.evergreen/find-python3.sh
@@ -318,29 +318,37 @@ find_python3() (
 # it will fall back to using find_python3 to return a suitable value.
 #
 ensure_python3() {
-  declare python_binary
-  declare test_path
-  python_binary="${DRIVERS_TOOLS_PYTHON:-}"
-  if [ -z "${python_binary}" ]; then
-    if [ "Windows_NT" == "${OS:-}" ]; then
-      test_path="/cygdrive/c/Python/Current/python.exe"
-    elif [ "$(uname -s)" != "Darwin" ]; then
-      test_path="/Library/Frameworks/Python.Framework/Versions/Current/bin/python3"
-    else
-      test_path="/opt/python/Current/bin/python3"
-    fi
-    if [ -f "${test_path}" ]; then
-      python_binary=${test_path}
-    fi
+  # Use "$DRIVERS_TOOLS_PYTHON".
+  if command -v "${DRIVERS_TOOLS_PYTHON:-}"; then
+    echo "Using Python binary ${DRIVERS_TOOLS_PYTHON:?}" >&2
+    echo "${DRIVERS_TOOLS_PYTHON:?}"
+    return
   fi
+
+  # Use Python Toolchain.
+  declare python_binary=""
+  case "${OSTYPE:?}" in
+  cygwin)
+    python_binary="/cygdrive/c/Python/Current/python.exe"
+    ;;
+  darwin*)
+    python_binary="/Library/Frameworks/Python.Framework/Versions/Current/bin/python3"
+    ;;
+  *)
+    python_binary="/opt/python/Current/bin/python3"
+    ;;
+  esac
+  if command -v "${python_binary:?}" >/dev/null; then
+    echo "Using Python binary ${python_binary:?}" >&2
+    echo "${python_binary:?}"
+    return
+  fi
+
+  # Use find_python3.
   {
-    if [ -z "${python_binary}" ]; then
-      echo "Finding Python3 binary..."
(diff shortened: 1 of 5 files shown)


.evergreen/find-python3.sh
```diff
@@ -314,11 +314,25 @@ find_python3() (
 # with `2>/dev/null` to silence these messages.
 #
 # If DRIVERS_TOOLS_PYTHON is set, it will return that value.  Otherwise
-# it will use find_python3 to return a suitable value.
+# it will look for the "Current" python in the python toolchain.  Finally,
+# it will fall back to using find_python3 to return a suitable value.
 #
 ensure_python3() {
```
eramongodb, line 320: Recommend using early return pattern instead to clearly distinguish the three separate cases for this function:

```bash
ensure_python3() {
  # Use "$DRIVERS_TOOLS_PYTHON".
  if command -v "${DRIVERS_TOOLS_PYTHON:-}"; then
    echo "Using Python binary ${DRIVERS_TOOLS_PYTHON:?}" >&2
    echo "${DRIVERS_TOOLS_PYTHON:?}"
    return
  fi

  # Use Python Toolchain.
  declare python_binary=""
  case "${OSTYPE:?}" in
  cygwin)
    python_binary="/cygdrive/c/Python/Current/python.exe"
    ;;
  darwin*)
    python_binary="/Library/Frameworks/Python.Framework/Versions/Current/bin/python3"
    ;;
  *)
    python_binary="/opt/python/Current/bin/python3"
    ;;
  esac
  if command -v "${python_binary:?}" >/dev/null; then
    echo "Using Python binary ${python_binary:?}" >&2
    echo "${python_binary:?}"
    return
  fi

  # Use find_python3.
  {
    echo "Finding Python3 binary..."
    python_binary="$(find_python3 2>/dev/null)" || return
    echo "Finding Python3 binary... done."
  } 1>&2
  echo "${python_binary:?}"
}
```

Additionally recommend using the Bash built-in env var `"$OSTYPE"` to distinguish the three primary distro types on Evergreen.

Additionally recommend using `command -v` rather than `-f` to account for the possibility of executable permissions issues.

.evergreen/install-cli.sh
```diff
@@ -10,6 +10,7 @@ fi
 
 if [ -z "${1:-}" ]; then
   echo "Must give a target directory!"
+  exit 1
```
eramongodb, line 13: Consider the following instead of `-z` + `exit`:

```bash
: "${1:?"must give a target directory!"}"
```

.evergreen/install-cli.sh
```diff
@@ -27,31 +28,55 @@ if [ "${DOCKER_RUNNING:-}" == "true" ]; then
...
 command -V uv # Ensure a working uv binary is present.
 
+# Ensure there is a venv available for backwards compatibility.
+if [ ! -d venv ]; then
+  uv venv venv
+fi
+
 # Store paths to binaries for use outside of current working directory.
-python_binary="$(which python)"
+python_binary="$(uv run python -c 'import sys;print(sys.executable)')"
```
eramongodb, line 79:
```suggestion
python_binary="$(uv run --no-project python -c 'import sys;print(sys.executable)')"
```

Avoid inadvertently detecting and using a project configuration at this point.

.evergreen/install-cli.sh
```diff
@@ -27,31 +28,55 @@ if [ "${DOCKER_RUNNING:-}" == "true" ]; then
...
+  fi
+fi
 
-if [ ! -d $SCRIPT_DIR/venv ]; then
+# If uv is still not available, we need a venv.
+if ! command -v uv >/dev/null; then
+  . ./venv-utils.sh
 
-  . ./find-python3.sh
+  if [ ! -d $SCRIPT_DIR/venv ]; then
```
eramongodb, line 50:
```suggestion
  if [ ! -d "$SCRIPT_DIR/venv" ]; then
```
ShellCheck SC2086: prevent globbing and word splitting.

.evergreen/install-cli.sh
```diff
@@ -27,31 +28,55 @@ if [ "${DOCKER_RUNNING:-}" == "true" ]; then
...
-  echo "Ensuring python binary... done."
+    . ./find-python3.sh
 
-  echo "Creating virtual environment 'venv'..."
-  venvcreate "${PYTHON:?}" venv
-  echo "Creating virtual environment 'venv'... done."
-else
-  venvactivate venv
+    echo "Ensuring python binary..."
+    PYTHON=$(ensure_python3 2>/dev/null)
```
eramongodb, line 55:
```suggestion
    PYTHON="$(ensure_python3 2>/dev/null)"
```
ShellCheck SC2086: prevent globbing and word splitting.

.evergreen/install-cli.sh
```diff
@@ -27,31 +28,55 @@ if [ "${DOCKER_RUNNING:-}" == "true" ]; then
...
+    _bin_path="/cygdrive/c/Python/Current"
+  elif [ "$(uname -s)" != "Darwin" ]; then
+    _bin_path="/Library/Frameworks/Python.Framework/Versions/Current"
+  else
+    _bin_path="/opt/python/Current"
+  fi
+  if [ -d "${_bin_path}" ]; then
+    export PATH="${_bin_path}:$PATH"
+  fi
+fi
```
eramongodb, line 44:
```suggestion
if ! command -v uv >/dev/null; then
  export PATH
  case "${OSTYPE:?}" in
  cygwin)
    PATH="/cygdrive/c/Python/Current/python.exe:${PATH:-}"
    ;;
  darwin*)
    PATH="/Library/Frameworks/Python.Framework/Versions/Current/bin/python3:${PATH:-}"
    ;;
  *)
    PATH="/opt/python/Current/bin/python3:${PATH:-}"
    ;;
  esac
fi
```

For symmetry with suggestions for `ensure_python3`.

eramongodb (changes requested):

.evergreen/find-python3.sh
blink1073: Done

.evergreen/install-cli.sh
blink1073: Done

.evergreen/install-cli.sh
```diff
@@ -27,33 +25,58 @@ if [ "${DOCKER_RUNNING:-}" == "true" ]; then
...
+# If uv is still not available, we need to install it into the venv.
 if ! command -v uv >/dev/null; then
   UV_UNMANAGED_INSTALL=1 python -m pip install -q --force-reinstall uv
 fi
 
 command -V uv # Ensure a working uv binary is present.
 
+# Ensure there is a venv available for backwards compatibility.
+if [ ! -d venv ]; then
+  uv venv venv
```
eramongodb, line 73: Just noting per [docs](https://docs.astral.sh/uv/reference/cli/#uv-venv):

> If a virtual environment exists at the target path, it will be removed and a new, empty virtual environment will be created.

This `uv venv venv` will blow away any pre-existing `venv` directory and replace it with a new one. If this is deliberate (just need _a_ venv, even if it's a fresh one), nothing to do.

eramongodb (approved):
Some minor feedback remaining; otherwise, LGTM.

PR #643, opened 2025-04-04
DRIVERS-3157 - Update Windows Python executable path

NoahStapp:


blink1073:
It doesn't appear to have worked: `No interpreter found for path `/cygdrive/c/python/Current/python.exe``.  Perhaps we need to use `"C:/python` instead of cygdrive?

blink1073:
It is a valid path on cygwin, but uv needs the c: version to run the executable.

NoahStapp:
> It is a valid path on cygwin, but uv needs the c: version to run the executable.

Is that a constraint of uv, or a restriction we've added?

blink1073:
It is a Windows restriction, the existing path was a bug.

blink1073 (approved):
LGTM!

PR #646, opened 2025-04-14
PYTHON-5307 Fix github app bootstrap and add test

blink1073:


.evergreen/github_app/utils.sh
```diff
@@ -13,6 +13,9 @@ function bootstrap() {
     bash ../install-node.sh
     source ../init-node-and-npm-env.sh
 
+    # Use the standard registry.
```
blink1073, line 16: This was the fix

NoahStapp (approved):

PR #653, opened 2025-05-12
DRIVERS-3188 Remove reliance on Evergreen instance profile credentials

blink1073:
- Remove internal usage of the instance profile creds
- Update relevant READMEs to reflect the requirement to assume a role

### DIFF
--- a/.evergreen/auth_aws/teardown.sh
+++ b/.evergreen/auth_aws/teardown.sh
@@ -5,7 +5,8 @@ set -o errexit
 SCRIPT_DIR=$(dirname ${BASH_SOURCE[0]})
 . $SCRIPT_DIR/../handle-paths.sh
 
-pushd $SCRIPT_DIR
+echo "Tearing down auth_aws..."
+pushd $SCRIPT_DIR >/dev/null
 
 # If we've gotten credentials, ensure the instance profile is set.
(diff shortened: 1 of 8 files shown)


.evergreen/auth_aws/teardown.sh
```diff
@@ -5,13 +5,16 @@ set -o errexit
...
-pushd $SCRIPT_DIR
+echo "Tearing down auth_aws..."
+pushd $SCRIPT_DIR >/dev/null
 
 # If we've gotten credentials, ensure the instance profile is set.
 if [ -f secrets-export.sh ]; then
   . ./activate-authawsvenv.sh
   source secrets-export.sh
-  python ./lib/aws_assign_instance_profile.py
+  python ./lib/aws_assign_instance_profile.py || true
```
NoahStapp, line 15: Should this call be fatal if it fails? Or since this is teardown, we don't care?

NoahStapp (changes requested):

.evergreen/auth_aws/teardown.sh
blink1073: The latter, we'll be removing this block entirely once the instance profile is removed.

NoahStapp (approved):

PR #662, opened 2025-06-26
DRIVERS-2884 Add mongoproxy to DET

prestonvasquez:
Add Go package for hosting a proxy to a MongoDB server that can intercept and forward messages for prose tests defined in [PR](https://github.com/mongodb/specifications/pull/1675). See DRIVERS-3183 for future use cases concerning unified spec tests.

Users will be expected to run the `start-mongoproxy.sh` script for non-auth, non-ssl, non-compressor tasks.

### DIFF
--- a/.evergreen/start-mongoproxy.sh
+++ b/.evergreen/start-mongoproxy.sh
@@ -1,0 +1,69 @@
+#!/usr/bin/env bash
+set -eux pipefail
+
+GOVERSION="${GOVERESION:-1.24}"
+GOPATH="${GOPATH:-$HOME/go}"
+
+# Detect OS
+OS="$(uname -s)"
+
+# If GOROOT is not set, determine it based on the OS and user-provided
+# GOVERSION.
+if [[ -z "${GOROOT:-}" ]]; then
+  case "$OS" in
+  Linux | Darwin)
+    GOROOT="/opt/golang/go${GOVERSION}"
+    ;;
+  MINGW* | MSYS* | CYGWIN*)
+    GOROOT="C:\\golang\\go${GOVERSION}"
+    ;;
+  *)
+    echo "unsupported OS: $OS"
+    exit 1
+    ;;
+  esac
+fi
+
+PATH="${GOROOT}/bin:${GOPATH}/bin:${PATH}"
+export GOROOT PATH
+
+echo "Using Go SDK at: $GOROOT (version: $GOVERSION)"
+
+test -x "${GOROOT}/bin/go" || {
+  echo "Go SDK not found at: $GOROOT"
+  exit 1
+}
+
+# Resolve this script’s dir, then go up one level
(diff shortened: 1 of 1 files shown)


.evergreen/start-mongoproxy.sh
```diff
@@ -0,0 +1,69 @@
...
+    ;;
+  MINGW* | MSYS* | CYGWIN*)
+    GOROOT="C:\\golang\\go${GOVERSION}"
+    ;;
+  *)
+    echo "unsupported OS: $OS"
+    exit 1
+    ;;
+  esac
+fi
```
baileympearson, line 25: Not sure if this is my particular Go environment being messed up (I installed Go following the instructions [here](https://go.dev/doc/install), macos arm64) but my go installation is in `/usr/local/go` and the script fails when I run it.

If I remove the GOROOT detection and set `GOROOT=/usr/bin/go`, this script works as expected for me

mongoproxy/README.md
```diff
@@ -0,0 +1,58 @@
+# mongoproxy
```
baileympearson, line 1: I don't feel strongly but we do have precedent of including mock servers in the `.evergreen` folder - should we move mongoproxy into `.evergreen`?  ex: `.evergreen/csfle` contains a lot of Python mock KMS servers

mongoproxy/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+Example:
+
+```
+{
+  "ping": 1,
+  "proxyTest": {
+    "actions": [
+      { "sendBytes": 1 },    // send only the first byte of the server response
+      { "delay": 200 },      // wait 200 milliseconds
```
baileympearson, line 49:
```suggestion
      { "delayMs": 200 },      // wait 200 milliseconds
```

baileympearson: Related to the above, I ran that example and saw no delay and no errors from the proxy.  Might be nice to log a warning if we encounter an unknown proxy action in the `actions` array?

baileympearson (changes requested):
@prestonvasquez No major comments here.  The prose tests, as written, would work in Node.

The proxy is pretty neat!  If other use cases arise, I might incorporate it into Node's CI in other places, it seems like it could be useful for testing other CMAP or SDAM scenarios.

.evergreen/start-mongoproxy.sh
prestonvasquez: @baileympearson I think most people will have the GOROOT installed on Darwin so I can make that a separate case that points to /user/local/go. It's common to globally export GOROOT, in which case this switch condition wouldn't execute. As you note, something like this should work:

```
GOROOT=/usr/local/go MONGODB_URI="mongodb://localhost:27017"  bash start-mongoproxy.sh
```

PR #665, opened 2025-07-02
Increase ephemeral storage and add diagnostics for k8s

vbabanin:
Our Gradle build began consuming more disk space to store dependencies, leading to pod evictions and failures due to insufficient ephemeral storage (exit code 137). This change increases the allocated storage to prevent these issues.

Additionally, to help troubleshoot future pod failures, this PR adds diagnostic logging. Previously, failures would return an exit code without meaningful context, making it harder to identify whether the issue was related to RAM or disk.

Verified by running these changes through a Java driver PR patch build:
[Example of a failed Java build with diagnostic output](https://spruce.mongodb.com/task/mongo_java_driver_test_oidc_k8s_variant_oidc_auth_test_k8s_gke_task_patch_b08097e1f5e5cf7d5c37ffa4b2932e64a2711ef6_68636bec72afa700072dcbf7_25_07_01_05_02_49/logs?execution=2)


blink1073 (approved):
LGTM, thanks!

PR #669, opened 2025-07-16
Increase GitHub Actions timeout

blink1073:
We were seeing timeouts on Windows, 5 minutes was not enough.


blink1073:
Evergreen failures will be fixed by #668

PR #670, opened 2025-07-17
Fix GitHub Actions timeouts

blink1073:
More robust follow up to #669

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,7 +20,7 @@ jobs:
   tests:
     name: "Tests"
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 10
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -20,7 +20,7 @@ jobs:
   tests:
     name: "Tests"
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 10
+    timeout-minutes: 5
```
caseyclements, line 23: We just changed this yesterday #669

caseyclements (changes requested):
Need to confirm reversion of timeout-minutes.

caseyclements (approved):
Read your PRs out-of-order. Good to go.

PR #672, opened 2025-07-25
DRIVERS-3223 Remove shasum check for gcloud

blink1073:


.evergreen/ensure-binary.sh
```diff
@@ -59,24 +59,22 @@ case $_NAME in
     popd
   ;;
   gcloud)
-    # NOTE: SHA_SUMs were copied from https://cloud.google.com/sdk/docs/install#installation_instructions
+    # NOTE: Google Cloud does not expose stable downloads with known shasums
+    # https://cloud.google.com/sdk/docs/install#installation_instructions lists
+    # shasums that change for every release.
```
eramongodb, line 64: 😢

eramongodb (approved):

eramongodb:
It looks like there are [versioned archives](https://cloud.google.com/sdk/docs/downloads-versioned-archives) which could be used instead of the "rapid release" archives, but there does not appear to be a definitive list of corresponding checksums (may need to be computed manually):

> Previous versions of the gcloud CLI are available in the [download archive](https://console.cloud.google.com/storage/browser/cloud-sdk-release) in Cloud Storage. They are directly accessible through `https://storage.googleapis.com/cloud-sdk-release/<archive_name>`.

PR #674, opened 2025-07-30
DRIVERS-3237 Add support for Python 3.14 and use Python 3.10 by default

blink1073:


blink1073:
Azure OIDC failure is pre-existing and tracked in [DRIVERS-3238](https://jira.mongodb.org/browse/DRIVERS-3238)

PR #676, opened 2025-08-06
DRIVERS-3249 fix pymongo compat with server version 4.0 and older

eramongodb:
Resolves DRIVERS-3249. Extends the changes made in https://github.com/mongodb-labs/drivers-evergreen-tools/pull/585 (to restore pymongo compatibility with MongoDB Server 3.6) to accomodate the recent pymongo 4.14 release, which dropped support for MongoDB Server 4.0 in https://github.com/mongodb/mongo-python-driver/pull/2353.


matthewdale (approved):
Looks good! 👍

PR #678, opened 2025-08-07
DRIVERS-3253 add support for uv.lock in install-cli.sh

eramongodb:
Resolves DRIVERS-3253. Followup to https://github.com/mongodb-labs/drivers-evergreen-tools/pull/585 and https://github.com/mongodb-labs/drivers-evergreen-tools/pull/676, motivated by [this comment](https://jira.mongodb.org/browse/DRIVERS-3249?focusedCommentId=7565273&focusedId=7565273&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-7565273) in DRIVERS-3249:

> We should have repeatable builds in so much as is possible.

Accordingly, this PR proposes creating a `uv.lock` file for each `pyproject.toml` in DET.

---

First, `requires-python` in `.evergreen/orchestration` is bumped to `3.9` for consistency with `.evergreen/pyproject.toml`, as otherwise:

```
× No solution found when resolving dependencies:
╰─▶ Because the requested Python version (>=3.8) does not satisfy Python>=3.9 and drivers-evergreen-tools==0.1.0 depends on Python>=3.9, we can conclude that drivers-evergreen-tools==0.1.0 cannot be used.
    And because only drivers-evergreen-tools==0.1.0 is available and your project depends on drivers-evergreen-tools, we can conclude that your project's requirements are unsatisfiable.
```

Second, a `uv.lock` file is created for each `pyproject.toml` currently in DET under `.evergreen` and `.evergreen/orchestration` using the command `uv lock`. Quoting uv [docs](https://docs.astral.sh/uv/concepts/projects/layout/#the-lockfile):

> `uv.lock` is a _universal_ or _cross-platform_ lockfile that captures the packages that would be installed across all possible Python markers such as operating system, architecture, and Python version.
>
> Unlike the `pyproject.toml`, which is used to specify the broad requirements of your project, the lockfile contains the exact resolved versions that are installed in the project environment. This file should be checked into version control, allowing for consistent and reproducible installations across machines.

There is nothing interesting in the lockfile for `.evergreen/pyproject.toml`, which is only used as a convenient means to install `mongodl.py` + related scripts as "tool executables". Regardless, this PR includes this file for consistency.

The lockfile for `.evergreen/orchestration/pyproject.toml` inherits the [dependencies of mongo_orchestration](https://github.com/mongodb-labs/mongo-orchestration/blob/fe08693ecccbe267207578c04a056950a158ac94/pyproject.toml#L37-L42) (its only direct dependency) and pins the package versions to the latest available (per the default "highest" resolution strategy). Notably, this includes `pymongo` (currently pinned to 4.14.0) as well as `certifi` (the `--with` in `install-cli.sh` is not necessary and therefore removed; this can be restored if it is required by some other downstream project/script beyond `.evergreen/orchestration/pyproject.toml`).

> [!NOTE]
> `uv` doesn't yet support `uv.lock` integration with `uv tool install` per [astral-uv/uv#5815](/astral-sh/uv/issues/5815). The ~~`uv run --quiet --frozen --isolated uv pip freeze`~~ `uv export` command is used as a workaround. This command ~~runs `uv sync --frozen` in an isolated virtual environment, which installs packages according to the project lockfile, then~~ outputs a `requirements.txt` file according to the list of installed packages. The output currently looks like (for `$MONGODB_VERSION >= 4.2`):
>
> ```
> # This file was autogenerated by uv via the following command:
> #    uv export --frozen --format requirements.txt -o uv-requirements.txt
> -e .
> ../
>     # via drivers-orchestration
> bottle==0.13.4 \
>     --hash=sha256:045684fbd2764eac9cdeb824861d1551d113e8b683d8d26e296898d3dd99a12e \
>     --hash=sha256:787e78327e12b227938de02248333d788cfe45987edca735f8f88e03472c3f47
>     # via mongo-orchestration
> certifi==2025.8.3 \
>     --hash=sha256:e564105f78ded564e3ae7c923924435e1daa7463faeab5bb932bc53ffae63407 \
>     --hash=sha256:f6c12493cfb1b06ba2ff328595af9350c65d6644968e5d3a2ffd78699af217a5
>     # via requests
> ...
> urllib3==2.5.0 \
>     --hash=sha256:3fc47733c7e419d4bc3f6b3dc2b4f890bb743906a30d56ba4a5bfa4bbff92760 \
>     --hash=sha256:e6b01673c0fa6a13e374b50871808eb3bf7046c4b125b216f6bf1cc604cff0dc
>     # via requests
> ```
>
> The requirements file is prefixed with `uv-` to avoid potentially conflicting with any pre-existing `requirements.txt` file in the target directory.

Third, to continue supporting backward compatibility with MongoDB Server versions that have been dropped by the latest PyMongo release (e.g. server versions 3.6 and 4.0), the [`--overrides`](https://docs.astral.sh/uv/reference/cli/#uv-tool-run) flag is added to the `uv tool install` command in `install-cli.sh` script. The overrides file may be specified by a new `DRIVERS_TOOLS_INSTALL_CLI_OVERRIDES` environment variable as appropriate for whichever project may be invoking the `install-cli.sh` script. This env var is used by `.evergreen/orchestration/setup.sh` to specify the `pymongo` constraints per `$MONGODB_VERSION`. The overrides file is named `uv-override-dependencies.txt` in reference to the [`[override-dependencies]`](https://docs.astral.sh/uv/reference/settings/#override-dependencies) uv project setting.

> [!NOTE]
> The `--with` flag cannot be used to override direct dependency requirements:
>
> ```
> × No solution found when resolving dependencies:
> ╰─▶ Because you require pymongo<4.14 and pymongo==4.14.0, we can conclude that your requirements are unsatisfiable.
> ```

---

The resulting behavior looks as follows given `MONGODB_VERSION=4.2` (from `.evergreen/orchestration`, with `--verbose`, focusing on the `pymongo` dependency, unrelated lines omitted for brevity):

```
DEBUG Adding direct dependency: pymongo>=4.14.0, <4.14.0+
DEBUG Adding transitive dependency for mongo-orchestration==0.12.0.dev0: pymongo>=4, <5
DEBUG Searching for a compatible version of pymongo (>=4.14.0, <4.14.0+)
DEBUG Selecting: pymongo==4.14.0 [compatible] (pymongo-4.14.0-cp313-cp313-manylinux_2_17_x86_64.manylinux2014_x86_64.whl)
 + pymongo==4.14.0
```

and with `MONGODB_VERSION=4.0` (which appends `pymongo<4.14` to the override dependencies file):

```
DEBUG Adding direct dependency: pymongo<4.14
DEBUG Adding transitive dependency for mongo-orchestration==0.12.0.dev0: pymongo<4.14
DEBUG Searching for a compatible version of pymongo (<4.14)
DEBUG Selecting: pymongo==4.13.2 [compatible] (pymongo-4.13.2-cp313-cp313-manylinux_2_17_x86_64.manylinux2014_x86_64.whl)
 + pymongo==4.13.2
```

### DIFF
--- a/.evergreen/install-cli.sh
+++ b/.evergreen/install-cli.sh
@@ -114,8 +114,7 @@ command -V uv
 uv --version
 
 # Workaround for https://github.com/astral-sh/uv/issues/5815.
-printf "" >|uv-requirements.txt
-uv run --quiet --frozen --isolated uv pip freeze >>uv-requirements.txt
+uv export --quiet --frozen --format requirements.txt -o uv-requirements.txt
 
 # Support overriding lockfile dependencies.
 if [[ ! -f "${DRIVERS_TOOLS_INSTALL_CLI_OVERRIDES:-}" ]]; then


blink1073:
> Possibly a uv version compatibility issue? 🤔

Yeah this might be stretching the lock file's cross-platform capability.

eramongodb:
> Possibly a uv version compatibility issue? 🤔

Looks like the `macos-14-arm64` distro is selecting a `uv` binary whose version (0.5.9, Dec 2024) is much older than other distros (0.7.0+, Apr 2025). [astral-sh/uv#10654](/astral-sh/uv/issues/10654) might be related. Regenerating the `uv.lock` file with `0.5.9` appears to have resolved the parsing issues while remaining compatible with newer `uv` versions.

eramongodb:
Alternatively, the `install-cli.sh` script could be updated to unconditionally prefer using the Python Toolchain provided `uv` binary (more likely to be up-to-date) over the system-provided `uv` binary by unconditionally adding the Python Toolchain bin directory to `PATH` (rather than conditioning it on the absence of a system-provided `uv` binary).

eramongodb:
I don't think we want to condition it on the `$CI` environment variable, since that would result in an inconsistency between test vs. real-world usage behavior.

Instead, I propose using the `uv` binary found via existing lookup routines to install _a more up-to-date_ uv binary via `uv tool install uv`. This will ensure a consistent `uv` binary version is used across all environments and is within our control. For starters, pinned the version to `uv~=0.8.0` to allow automatic patch releases updates.

blink1073:
Perhaps we should add a Troubleshooting section to the Readme, and say "if you see this error, here is where to apply the fix".

blink1073:
> Perhaps we should add a Troubleshooting section to the Readme, and say "if you see this error, here is where to apply the fix".

I'm concerned about the discoverability of knowing to make this change when PyMongo drops support for a MongoDB version.

eramongodb:
> I'm concerned about the discoverability of knowing to make this change when PyMongo drops support for a MongoDB version.

Added a section to the README documenting the PyMongo dependency and update instructions.

blink1073 (approved):
Excellent, thank you!

blink1073:
`bash .evergreen/prepare-env-and-resources.sh` runs fine on my local machine, adding a trace might help.

blink1073:
I echoed the content of `uv-requirements.txt` in this [patch](https://spruce.mongodb.com/task/drivers_tools_tests_os_requires_70__os_requires_70~macos_14_arm64_auth~auth_ssl~ssl_test_latest_standalone_patch_61eacf1d1f0424d2a8c0bc471b0da50b162c5e42_68a61dfd5ab3410007e86023_25_08_20_19_12_05/logs?execution=0):

```
[2025/08/20 14:16:11.964] certifi==2025.7.14
[2025/08/20 14:16:11.964] pip==22.3.1
[2025/08/20 14:16:11.964] pip==25.1.1
[2025/08/20 14:16:11.964] rust-just==1.42.4
[2025/08/20 14:16:11.964] setuptools==65.5.0
[2025/08/20 14:16:11.964] setuptools==80.9.0
[2025/08/20 14:16:11.964] uv==0.8.3
```

eramongodb:
Solved: it's a Python binary compatibility problem due to the `uv run uv pip freeze` pattern. Successful tasks [find](https://parsley.mongodb.com/evergreen/drivers_tools_tests_os_requires_70__os_requires_70~macos_14_arm64_auth~auth_ssl~ssl_test_cli_partial_patch_ee516797546cd982872f77ff509051fdac5df7d8_68a5f5965c4c0e00070bf895_25_08_20_16_19_36/0/task?bookmarks=0,3490&shareLine=173) Python 3.13, whereas failing tasks [find](https://parsley.mongodb.com/evergreen/drivers_tools_tests_os_requires_70__os_requires_70~macos_14_arm64_auth~auth_ssl~ssl_test_cli_partial_patch_ee516797546cd982872f77ff509051fdac5df7d8_68a5f5965c4c0e00070bf895_25_08_20_16_19_36/1/task?bookmarks=0,470&shareLine=173) Python 3.11. It seems the `uv pip freeze` command may be including seed packages required by the isolated virtual environment in its output.

It turns out a more suitable command for this purpose which does not require `pip` or an isolated virtual environment is [`uv export`](https://docs.astral.sh/uv/reference/cli/#uv-export). Even when it [finds](https://parsley.mongodb.com/evergreen/drivers_tools_tests_os_requires_70__os_requires_70~macos_14_arm64_auth~auth_ssl~ssl_test_cli_partial_patch_ee516797546cd982872f77ff509051fdac5df7d8_68a61902e979cc00072eb87d_25_08_20_18_50_43/1/task?bookmarks=0,3619&shareLine=2929) Python 3.11, `pip` and `setuptools` are not included [in the requirements](https://parsley.mongodb.com/evergreen/drivers_tools_tests_os_requires_70__os_requires_70~macos_14_arm64_auth~auth_ssl~ssl_test_cli_partial_patch_ee516797546cd982872f77ff509051fdac5df7d8_68a61902e979cc00072eb87d_25_08_20_18_50_43/1/task?bookmarks=0,3619&shareLine=2638).

PR description has been updated accordingly.

blink1073 (approved):
Thanks again!

eramongodb:
> It is because we have `auth_aws` as part of the workspace in `pyproject.toml`.

I don't this is correct. The error suggests `uv tool install .` is being run in the `.evergreen/auth_aws` directory (which has no `pyproject.toml`). `install-cli.sh` should be installing `.evergreen` (the DET CLI tools) or `.evergreen/orchestration` (for `drivers-orchestration`), not `.evergreen/auth_aws`. `install-cli.sh` is invoked from the `.evergreen` directory in all cases except `.evergreen/orchestration/setup.sh`, which invokes it twice: once for `.evergreen` and once more with `.evergreen/orchestration`. Where/how is running `make run-server` resulting in `install-cli.sh` being invoked with a target directory of `.evergreen/auth_aws`?

blink1073:
No, uv is attempting to build a project in the workspace, and failing to do so, because it has no pyproject.toml file.  As to why it occurs when invoking one script and not the other, I'm not sure, but I first noticed this when running our server start script from pymongo, which is also broken in the same way.

ShaneHarvey:
I’ll defer to Steve’s review.

PR #682, opened 2025-08-14
move `-v` flag from `docker build` to `docker run`

kevinAlbs:
To fix an observed error when running `start_local_server.sh` with `AWS_PROFILE` exported:
```bash
# AWS_PROFILE set previously
export AWS_PROFILE        
.evergreen/auth_oidc/start_local_server.sh
```

Results in:
```
unknown shorthand flag: 'v' in -v
```


blink1073 (approved):
Thanks!

PR #688, opened 2025-09-02
Remove symlinked uv executables and add to gitignore

blink1073:
Follow up to #685 where I accidentally committed these files.  They are created locally due to #678.


GromNaN (approved):

blink1073:
> Does that fixes this [failing job](https://github.com/doctrine/mongodb-odm/actions/runs/17412704044/job/49433597419?pr=2813)?

I believe it will, yes

PR #690, opened 2025-09-16
always stop mongo orchestration in drivers-orchestarion.py

baileympearson:
Consecutive runs of drivers-orchestation.py fail after the first because mongo-orchestration is already running.  This PR updates drivers-orchestration to stop mongo-orchestration, if it is running, before attempting to launch it again.

### DIFF
--- a/.evergreen/orchestration/drivers_orchestration.py
+++ b/.evergreen/orchestration/drivers_orchestration.py
@@ -477,6 +477,8 @@ def run(opts):
         orch_file = Path(mo_home / "config.json")
         orch_file.write_text(json.dumps(data, indent=2))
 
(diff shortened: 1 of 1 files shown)


blink1073:
Stopping is already part of [stop](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/d7a7337b384392a09fbe7fc80a7244e6f1226c18/.evergreen/orchestration/drivers_orchestration.py#L553), we could remove the conditional there.

blink1073:
The change looks good to me to pick up the case where docker had been used to start the cluster.  Was that the issue you were facing?  If not, stop relies on a server.pid file being there one way or the other.

PR #790, opened 2026-06-04
DRIVERS-2920 add KMS HTTP proxy server

mdb-ad:
DRIVERS-2920 will add HTTP proxy support. This PR adds a Python HTTP proxy for testing. Besides the CONNECT method there's also a GET /metrics so that drivers can verify that the proxy observed a connection

### DIFF
--- a/.evergreen/csfle/kms_http_proxy.py
+++ b/.evergreen/csfle/kms_http_proxy.py
@@ -2,10 +2,10 @@
 """Minimal HTTPS CONNECT proxy with a /metrics endpoint.
 
 Plain HTTP (no inbound TLS):
-    python3 kms_http_proxy.py [--host 127.0.0.1] [--port 8080]
+    python3 kms_http_proxy.py [--host 127.0.0.1] [--port 9004]
 
-    $ curl --proxy 127.0.0.1:8080 -sS -o /dev/null https://example.com
-    $ curl http://127.0.0.1:8080/metrics
+    $ curl --proxy 127.0.0.1:9004 -sS -o /dev/null https://example.com
+    $ curl http://127.0.0.1:9004/metrics
 
 TLS on the inbound connection (clients must connect over HTTPS):
     python3 kms_http_proxy.py --port 8443 --ca_file ca.pem --cert_file server.pem
@@ -160,9 +160,14 @@ def _tunnel(a: socket.socket, b: socket.socket) -> None:
 def main() -> int:
     p = argparse.ArgumentParser()
     p.add_argument("--host", default="127.0.0.1")
-    p.add_argument("--port", type=int, default=8080)
+    p.add_argument("--port", type=int, default=9004)
     p.add_argument("--ca_file", type=str, default=None, help="TLS CA PEM file")
-    p.add_argument("--cert_file", type=str, default=None, help="TLS server PEM file (required with --ca_file)")
+    p.add_argument(
+        "--cert_file",
+        type=str,
+        default=None,
+        help="TLS server PEM file (required with --ca_file)",
+    )
     args = p.parse_args()
 
     if bool(args.ca_file) != bool(args.cert_file):
--- a/.evergreen/csfle/kms_http_proxy.py
+++ b/.evergreen/csfle/kms_http_proxy.py
@@ -4,10 +4,12 @@
 Run:
     python3 kms_http_proxy.py [--host 127.0.0.1] [--port 8080]
 
-Then point the test at it:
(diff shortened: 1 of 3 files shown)


.evergreen/csfle/kms_http_proxy.py
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env python3
+"""Minimal HTTPS CONNECT proxy with a /metrics endpoint.
+
+Run:
+    python3 kms_http_proxy.py [--host 127.0.0.1] [--port 8080]
+
+Then point the test at it:
+    export MONGOC_TEST_KMS_PROXY_HOST=127.0.0.1
+    export MONGOC_TEST_KMS_PROXY_PORT=8080
+    ./test-libmongoc -l '/client_side_encryption/kms/connect_callback/*'
```
kevinAlbs: Suggest showing example use with `curl` rather than C driver test runner.

```suggestion
To test:

    $ curl --proxy 127.0.0.1:8080 -sS -o /dev/null https://example.com
    $ curl http://127.0.0.1:8080/metrics
    connect_count 1
    connect_target example.com:443
```

.evergreen/csfle/kms_http_proxy.py
```diff
@@ -0,0 +1,153 @@
...
+
+Behavior:
+  - CONNECT host:port HTTP/1.1  -> opens a TCP tunnel, returns 200, then
+                                   blindly proxies bytes both directions.
+  - GET /metrics                -> returns "connect_count <N>\n" so callers
+                                   can verify a CONNECT was observed.
+  - Any other request           -> 405.
+
+No auth, no TLS on the inbound side, no concurrency limits. This is for
```
kevinAlbs: I expect this proxy needs to support both TLS and no-TLS (corresponding to `HTTP_PROXY` / `HTTPS_PROXY`) to meet the [scope](https://jira.mongodb.org/browse/WRITING-37418) test goal: "Connect to proxy over HTTP and HTTPS."

In particular, I want drivers to test both TLS and no-TLS connections to the proxy because that may influence the KMS callback return type.

For example: if the C driver only needs to connect to the proxy with no-TLS, the callback could return an `int` socket descriptor. But to support returning a socket with TLS, the return needs to return an abstraction like `mongoc_stream_t`.

Supporting TLS could use existing certificates in this repo, similar to [kms_http_server.py](https://github.com/mdb-ad/drivers-evergreen-tools/blob/cd5c8829dfcbbf7959ba3a38229151261caf0493/.evergreen/csfle/kms_http_server.py#L254-L258).

.evergreen/csfle/kms_http_proxy.py
```diff
@@ -138,11 +140,27 @@ def main() -> int:
     p = argparse.ArgumentParser()
     p.add_argument("--host", default="127.0.0.1")
     p.add_argument("--port", type=int, default=8080)
+    p.add_argument("--ca_file", type=str, default=None, help="TLS CA PEM file")
+    p.add_argument("--cert_file", type=str, default=None, help="TLS server PEM file (required with --ca_file)")
```
kevinAlbs: Run `pre-commit run --all-files` to fix lint task.

kevinAlbs (approved):
LGTM with a lint fix.

.evergreen/csfle/kms_http_proxy.py
```diff
@@ -0,0 +1,173 @@
...
+            try:
+                peer.sendall(data)
+            except OSError:
+                return
+
+
+def main() -> int:
+    p = argparse.ArgumentParser()
+    p.add_argument("--host", default="127.0.0.1")
+    p.add_argument("--port", type=int, default=8080)
```
blink1073: This default is the same as the mock Azure IMDS server, suggest using a different port to avoid conflicts.

.evergreen/csfle/kms_http_proxy.py
```diff
@@ -0,0 +1,173 @@
...
+        _connect_count += 1
+        _connect_targets.append(target)
+
+
+def _snapshot_metrics() -> tuple[int, list[str]]:
+    with _counter_lock:
+        return _connect_count, list(_connect_targets)
+
+
+class ProxyHandler(BaseHTTPRequestHandler):
```
blink1073, line 61: I suggest adding a `/reset` endpoint to resent the counter, rather than requiring the server to be re-started between runs.

blink1073 (changes requested):

blink1073 (approved):
LGTM

PR #798, opened 2026-07-02
DRIVERS-3558: skip mongodb-runner on unsupported platforms

blink1073:
DRIVERS-3558

## Summary

Platforms that can't run the current mongodb-runner now fall back to mongo-orchestration automatically, rather than relying on npm overrides to work around version incompatibilities.

## Changes in this PR

- Replaces the npm overrides workaround with a platform capability check that falls back to mongo-orchestration when unsupported, checking the actual Node version rather than only distro name so the check generalizes beyond RHEL7.
- Hardens the install-binaries test script (Windows-safe invocation, clearer failure on a missing Python binary, and always exercising a fresh install).

## Test Plan

- Updated `.evergreen/tests/test-install-binaries.sh` to exercise the real install/support-check code path.
- Ran `pre-commit run` on all changed files (shellcheck, ruff, ruff-format, etc.) — all passing.
- Verified locally that the platform check and install-and-run path work end-to-end.

### Checklist for Author

### Checklist for Reviewer

### DIFF
--- a/.evergreen/tests/test-install-binaries.sh
+++ b/.evergreen/tests/test-install-binaries.sh
@@ -6,28 +6,34 @@ SCRIPT_DIR=$(dirname ${BASH_SOURCE[0]})
 
 pushd $SCRIPT_DIR/..
 
+# The system python3 may be too old (or missing); find a suitable one.
+. ./find-python3.sh
+PYTHON_BINARY=$(ensure_python3 2>/dev/null)
+PATH="$(dirname $PYTHON_BINARY):$PATH"
+
 ./install-node.sh
 . ./init-node-and-npm-env.sh
-MR_INSTALL_DIR=$(mktemp -d)
-case "$(uname -s)" in
-  CYGWIN*) MR_INSTALL_DIR=$(cygpath -m "$MR_INSTALL_DIR") ;;
-esac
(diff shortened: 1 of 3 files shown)


eramongodb (approved):

eramongodb:
Just realized this PR was changed to draft: is this still being worked on?

blink1073:
I had made some follow-up changes based on a claude PR review, then I was waiting for builds to pass.  We're good to go now.

.evergreen/tests/test-install-binaries.sh
```diff
@@ -1,33 +1,49 @@
...
+set -eu -o pipefail
 
 SCRIPT_DIR=$(dirname ${BASH_SOURCE[0]})
 . $SCRIPT_DIR/../handle-paths.sh
 
 pushd $SCRIPT_DIR/..
 
+# The system python3 may be too old (or missing); find a suitable one.
+. ./find-python3.sh
+PYTHON_BINARY=$(ensure_python3 2>/dev/null)
```
eramongodb, line 11: Aside: would be nice to migrate this to `uv` eventually, but fine with leaving out-of-scope for this PR.

eramongodb (approved):

PR #818, opened 2026-07-31
DRIVERS-3564: Reuse an existing --user uv install instead of reinstalling it

blink1073:
DRIVERS-3564

## Summary

`ensure_uv` reinstalls uv on hosts where the `pip install --user` script
directory isn't on the default `PATH`, because the `PATH` it sets doesn't
outlive the shell. On rhel7 a single `test-mongodb-runner-partial` task ran
`pip install --user uv` 15 times, roughly 15–20s of a 3m11s task.

## Changes in this PR

`ensure_uv` now looks for an already-installed uv in that directory before
falling back to installing one.

## Test Plan

No unit tests exist for these shell helpers, so each branch was exercised
directly, reproducing the rhel7 condition locally:

- uv present but directory off `PATH`: reused, no install
- uv already on `PATH`: unchanged
- called repeatedl: `PATH` does not grow
- uv genuinely absent: still installs, then fails with the same actionable error

Confirmed in CI on the rhel7 variant: `installing with ... pip install --user uv`
now appears **once** per task instead of 15 times, with the 17 later invocations
resolving the existing install instead. The task ran in 2m40s vs 3m11s before.

### Checklist for Author

### Checklist for Reviewer

### DIFF
--- a/.evergreen/ensure-uv.sh
+++ b/.evergreen/ensure-uv.sh
@@ -70,19 +70,13 @@ _ensure_uv_add_user_base() {
   user_base="$("$1" -m site --user-base 2>/dev/null)" || return 0
   [ -n "$user_base" ] || return 0
 
-  # Build the prefix first and prepend it once, so "bin" keeps precedence over
-  # "Scripts" as it would with a single assignment; prepending in a loop would
-  # reverse them.
-  declare _prefix="" _dir
-  for _dir in "$user_base/bin" "$user_base/Scripts"; do
-    case ":${PATH:-}:" in
-    *":${_dir}:"*) ;;
-    *) _prefix="${_prefix}${_dir}:" ;;
-    esac
-  done
-  if [ -n "$_prefix" ]; then
-    export PATH="${_prefix}${PATH:-}"
-  fi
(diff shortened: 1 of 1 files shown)


.evergreen/ensure-uv.sh
```diff
@@ -57,6 +57,28 @@ _ensure_uv_scope_paths() {
...
+#
+# Prepend $1's `pip install --user` script directory to PATH, if not already
+# there. That directory is version/platform specific (~/.local/bin on Linux,
+# ~/Library/Python/X.Y/bin on macOS, %APPDATA%\Python\PythonXY\Scripts on
+# Windows), so ask the interpreter rather than assuming.
+#
+# A no-op if the interpreter cannot report it. Not meant to be called directly.
+_ensure_uv_add_user_base() {
+  declare user_base
+  user_base="$("$1" -m site --user-base 2>/dev/null)" || return 0
```
eramongodb:
```suggestion
  user_base="$("${1:?}" -m site --user-base 2>/dev/null)" || return 0
```
Enforce never null-or-empty precondition.

eramongodb (approved):
Minor suggestion; otherwise, LGTM.

PR #825, opened 2026-08-05
DRIVERS-3564: Install uv into a virtual environment when there is no system pip

blink1073:
DRIVERS-3564

## Summary

`ensure_uv` used `pip install --user`, which needs a system pip. The remote KMS
VMs have none: they install `python3-venv`, not `python3-pip`, and Debian and
Ubuntu disable `ensurepip` for the system python. The gcpkms and azurekms tasks
failed with `No module named pip`.

[4c63815](https://github.com/mongodb-labs/drivers-evergreen-tools/commit/4c63815912cb0a4ff7bc877c173d0deab39690c5) added `python3-pip` to the
provisioning scripts, but those come from the driver's pinned
`drivers-evergreen-tools`, so it only helps once a driver bumps its pin. This
fixes the side that runs on the VM, so no driver has to act.

## Changes in this PR

[97b2779](https://github.com/mongodb-labs/drivers-evergreen-tools/commit/97b2779) falls back to installing uv into a virtual environment,
which brings its own pip and needs only the stdlib module those VMs have.

Both install paths are kept, because the fleet is split in both directions:
Evergreen's debian11 images have `python3-pip` but no `python3-venv`, so a venv
cannot be built there, while the KMS VMs are the reverse. Neither path alone
covers both.

## Test Plan

`test-remote-kms-provisioning.sh` gained cases for both shapes: a host with
`python3-venv` and no `python3-pip` (debian:11 and ubuntu:20.04, the latter also
covering a bundled pip predating PEP 600), and a host with `python3-pip` and no
`python3-venv` (debian:11). Each asserts uv ends up installed, and disabling
either install path makes the corresponding case fail.

Verified on real GCE and Azure VMs with the two node driver tasks that originally
failed, with the remote clone temporarily pointed at this branch, since the VM
otherwise clones the default branch and would not pick the change up:
https://spruce.corp.mongodb.com/version/6a7338dda768b70007e5601d

A follow-up PR will have remote VMs unpack an archive of the host's checkout
rather than cloning the default branch, which removes the need for that override.

### Checklist for Author

### Checklist for Reviewer

### DIFF
--- a/.evergreen/ensure-uv.sh
+++ b/.evergreen/ensure-uv.sh
@@ -57,26 +57,33 @@ _ensure_uv_scope_paths() {
   export UV_TOOL_DIR="${DRIVERS_TOOLS}/.local/uv-tool"
 }
 
-# _ensure_uv_add_user_base (internal)
+# _ensure_uv_add_path (internal)
 #
-# Prepend $1's `pip install --user` script directory to PATH, if not already
-# there. That directory is version/platform specific (~/.local/bin on Linux,
-# ~/Library/Python/X.Y/bin on macOS, %APPDATA%\Python\PythonXY\Scripts on
-# Windows), so ask the interpreter rather than assuming.
+# Prepend $1 to PATH unless it is already there, which keeps repeated ensure_uv
+# calls from growing PATH without end. Not meant to be called directly.
+_ensure_uv_add_path() {
+  declare dir="${1:?}"
+  case ":${PATH:-}:" in
+  *":$dir:"*) ;;
+  *) export PATH="$dir:${PATH:-}" ;;
+  esac
+}
+
+# _ensure_uv_add_user_bin (internal)
+#
+# Put $1's `pip install --user` script directory on PATH. That directory is
+# version and platform specific (~/.local/bin on Linux, ~/Library/Python/X.Y/bin
+# on macOS, %APPDATA%\Python\PythonXY\Scripts on Windows), so ask the interpreter
+# rather than assuming. Only one of bin/Scripts exists on any given platform, so
+# adding both is harmless.
 #
 # A no-op if the interpreter cannot report it. Not meant to be called directly.
-_ensure_uv_add_user_base() {
-  declare user_base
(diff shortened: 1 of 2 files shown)


blink1073:
cc @PavelSafronov

.evergreen/tests/test-remote-kms-provisioning.sh
PavelSafronov, line 1: `test_no_system_pip` and `test_no_venv_module` are great, but can you also add a test that runs `ensure_uv` inside an active venv? That's the Node OIDC tests failure mode: pip is present, but `pip install --user` is refused in a venv. It's also the only case that exercises the fallback's `python3 -m venv` from a venv interpreter. This would guard the venv-in-venv from regressing in the future.

.evergreen/ensure-uv.sh
```diff
@@ -140,38 +160,64 @@ ensure_uv() {
...
+    if "$py" -m pip --version >/dev/null 2>&1; then
+      echo "uv not found; installing it with '$py -m pip install --user uv'..." >&2
+
+      # PIP_BREAK_SYSTEM_PACKAGES bypasses PEP 668's externally-managed-environment
+      # guard, which Debian and Ubuntu enable by default. Safe here: a --user
+      # install does not touch system site-packages.
+      #
+      # Upgrade pip first, since one predating PEP 600 (e.g. 20.0.2 on Ubuntu
+      # 20.04) mis-resolves uv's wheel tags. A fast no-op when already current.
+      PIP_BREAK_SYSTEM_PACKAGES=1 "$py" -m pip install --user -q --upgrade pip || true
```
PavelSafronov: `|| true` leads to opaque failures. Can you surface the last pip/venv errors?

.evergreen/tests/test-remote-kms-provisioning.sh
blink1073: Done

.evergreen/ensure-uv.sh
blink1073: Done

PavelSafronov (approved):
Changes look great, this should fix the failing Node Drivers CI builds, approving.

PR #834, opened 2026-08-27
DRIVERS-3626: Lock the mongodb-runner install and pin its driver per server version

blink1073:
DRIVERS-3626

## Summary

`mongodb` 7.6.0 raised the Node driver's `minWireVersion` to 9, which breaks the handshake with every MongoDB 4.2 server. The `mongodb-runner` install had no lockfile, so that upstream release reached CI on its own: a mainline revision that passed on 2026-08-22 failed on 2026-08-24, with no change to this repo in between.

The driver version now comes from a committed lockfile, so it changes only when someone edits that file in a pull request. 4.2 gets its own pin at `mongodb` 7.5.0, the last release before the `minWireVersion` change.

Replaces the #833 stop-gap and removes its version-gate entry.

## Changes in this PR

- Resolve the `mongodb-runner` install from committed lockfiles using `npm ci`.
- Pin the driver per server version, holding 4.2 at the last release that speaks wire version 8.
- Reject an inconsistent pin set at install time instead of degrading to `mongo-orchestration`.
- Route 4.2 back through `mongodb-runner`, removing the stop-gap's gate entry.
- Track the pins in Dependabot, with security-only updates for held-back pins.

## Test Plan

- `run-mongodb.sh start --version 4.2` came up through pin set `mongodb-runner-6.8.2-mongodb-7.5.0`, reporting `maxWireVersion: 8` and `ok:1` from `ping`.
- Asserted the pin logic directly: an invalid pin raises on a cache hit, an `overrides`/lockfile mismatch raises, and traversal-shaped versions fall through to the default pin.
- `make lint` passes.

The matrix and the other server versions only run in Evergreen, so a patch build is the real gate.

### Checklist for Author

- [x] Is all relevant documentation (README or docstring) updated? Adds `.evergreen/orchestration/mongodb-runner/README.md`.

### Checklist for Reviewer

### DIFF
--- a/.evergreen/orchestration/drivers_orchestration.py
+++ b/.evergreen/orchestration/drivers_orchestration.py
@@ -567,16 +567,16 @@ def run(opts):
     if opts.local_atlas:
         opts.mongodb_runner = False
 
-    # mongodb-runner drives the server with the Node driver, which dropped 4.2 in
-    # mongodb 7.6.0. mongodb-runner asks for ^7.2.0, so it can no longer reach a
-    # 4.2 server at all: the handshake fails on wire version 8.
-    if opts.mongodb_runner and version in ("3.6", "4.0", "4.2"):
+    # A bare match on version misses "4.0.28" or "v4.0", which are exactly as
+    # unreachable as "4.0" itself, so compare the X.Y prefix instead.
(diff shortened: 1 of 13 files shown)


PavelSafronov (approved):
Looks good, no blockers.

PR #842, opened 2026-09-11
PYTHON-6005 & DRIVERS-3564 Fix ensure_uv Python selection and active-venv install

blink1073:
PYTHON-6005
DRIVERS-3564

## Summary

Bug fix and cleanup pass for ensure_uv.  The bug for PYTHON-6005 was that we were selecting the system python3 on `tests-rhel82-arm64`, which is Python 3.6.
We also further improve the use case for Node with OIDC by installing uv into an active venv rather than making another temp venv.
There are also some new tests and improvements to the Dockerfiles.

## Changes in this PR

- Update the logic to prefer the latest mongodb toolchain python over the system python unconditionall.
- Install uv into an active venv (upgrading pip first) rather than using `--user`, which pip refuses inside a venv.
- Add a per-PR `test-ensure-uv.sh` regression test for the active-venv and pip-without-venv cases. This is the bulk of the diff.
- Add a `tests-rhel82-arm64` buildvariant (with step to clear any uv added by a previous run), proving the toolchain is chosen when the system python3 is too old.
- Drop the obsolete fallback-venv assertion in `test-remote-kms-provisioning.sh`.
- Update the Dockerfiles to make sure pip is available, removing the need for the venv fallback pass there.

## Test Plan

- `make lint` clean; `config.yml` parses; `bash -n` on changed scripts; `test-ensure-uv.sh` passes; drivers-tools Evergreen patch green.
- Passing [build](https://spruce.corp.mongodb.com/version/6aa48e11a2c38400076f8daa/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) with Node OIDC to verify that the active venv case still works.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -568,6 +568,13 @@ functions:
         binary: bash
         args: [src/.evergreen/tests/test-drivers-tools-archive.sh]
 
+  "run ensure uv test":
+    - command: subprocess.exec
+      type: test
+      params:
+        binary: bash
+        args: [src/.evergreen/tests/test-ensure-uv.sh]
+
   "run mongodb runner test partial":
     - command: subprocess.exec
       type: test
@@ -624,6 +631,16 @@ functions:
           rm -rf ${DRIVERS_TOOLS} || true
 
 pre:
+  # Inline rather than a script because this runs before git.get_project, so the
+  # checkout that would hold the script does not exist yet.
+  - command: shell.exec
+    params:
+      shell: bash
+      script: |
+        if [ "${clear_persisted_uv|}" = "true" ]; then
+          echo "Removing any uv an earlier task left behind, so ensure_uv has to pick an interpreter."
+          rm -f "$HOME/.local/bin/uv"
+        fi
   - func: "fetch and prepare resources"
   - func: "assume role"
 
@@ -1053,6 +1070,11 @@ tasks:
       commands:
         - func: "run drivers tools archive test"
 
+    - name: "test-ensure-uv"
+      tags: ["pr"]
+      commands:
+        - func: "run ensure uv test"
+
     - name: "test-legacy-versions"
       commands:
         - func: "run legacy versions test"
@@ -1466,6 +1488,20 @@ buildvariants:
     - ".local-atlas"
     - ".github-app"
 
+# RHEL 8.2 arm64 is the host PYTHON-6005 was filed for: its system python3 is
+# 3.6, which uv publishes no distribution for, and that Evergreen image does not
+# ship uv. clear_persisted_uv drops any uv an earlier task installed under the
+# home directory, so ensure_uv has to select an interpreter and only the
+# toolchain can install one. Reversing that precedence fails this variant.
+- name: tests-rhel82-arm64
+  display_name: "RHEL 8.2 ARM64"
+  run_on:
+    - rhel82-arm64-small
+  expansions:
+    clear_persisted_uv: "true"
+  tasks:
+    - name: "test-latest-standalone"
+
 - name: tests-rhel
   display_name: RHEL Tests
   run_on:
@@ -1524,6 +1560,7 @@ buildvariants:
           - "test-install-binaries"
           - "test-csfle"
           - "test-drivers-tools-archive"
+          - "test-ensure-uv"
           - "test-cli-full"
           - "test-mongodb-runner-full"
           - "test-happy-eyeballs"
--- a/.evergreen/ensure-uv.sh
+++ b/.evergreen/ensure-uv.sh
@@ -186,11 +186,8 @@ ensure_uv() {
   declare venv_dir="${TMPDIR:-/tmp}"
   venv_dir="${venv_dir%/}/drivers-tools-uv-venv"
 
-  # Prefer the MongoDB toolchain's python3, which is modern, over the system one.
-  # On the rhel82-arm64-small Evergreen image the system python3 is 3.6, which uv
-  # publishes no distribution for, so a host that has the toolchain would
-  # otherwise fail to install uv. Legacy hosts (rhel7) have no python3 on PATH at
-  # all, only an ancient Python 2.
+  # Prefer the MongoDB toolchain's python3: the system one is 3.6 on
+  # rhel82-arm64-small, and rhel7 has no python3 on PATH at all.
   declare py="" toolchain_py
   toolchain_py="$(compgen -G '/opt/mongodbtoolchain/v*/bin/python3' | sort -V | tail -n1)" || true
   if [ -n "$toolchain_py" ] && [ -x "$toolchain_py" ]; then
--- a/.evergreen/ensure-uv.sh
+++ b/.evergreen/ensure-uv.sh
@@ -191,9 +191,14 @@ ensure_uv() {
   # python3, which is modern, over the system one: it is 3.6 on
   # rhel82-arm64-small, and rhel7 has no python3 on PATH at all.
   declare py="" toolchain_py
-  if [ -n "${VIRTUAL_ENV:-}" ] && [ -x "$VIRTUAL_ENV/bin/python" ]; then
-    py="$VIRTUAL_ENV/bin/python"
-  else
+  if [ -n "${VIRTUAL_ENV:-}" ]; then
+    if [ -x "$VIRTUAL_ENV/bin/python" ]; then
+      py="$VIRTUAL_ENV/bin/python"
+    elif [ -x "$VIRTUAL_ENV/Scripts/python.exe" ]; then
+      py="$VIRTUAL_ENV/Scripts/python.exe"
+    fi
+  fi
+  if [ -z "$py" ]; then
     toolchain_py="$(compgen -G '/opt/mongodbtoolchain/v*/bin/python3' | sort -V | tail -n1)" || true
     if [ -n "$toolchain_py" ] && [ -x "$toolchain_py" ]; then
       py="$toolchain_py"
--- a/.evergreen/ensure-uv.sh
+++ b/.evergreen/ensure-uv.sh
@@ -186,16 +186,22 @@ ensure_uv() {
   declare venv_dir="${TMPDIR:-/tmp}"
   venv_dir="${venv_dir%/}/drivers-tools-uv-venv"
 
-  # Prefer the MongoDB toolchain's python3: the system one is 3.6 on
+  # Use the active venv's interpreter so uv installs into it; see the in-venv
+  # branch of _ensure_uv_install. Otherwise prefer the MongoDB toolchain's
+  # python3, which is modern, over the system one: it is 3.6 on
   # rhel82-arm64-small, and rhel7 has no python3 on PATH at all.
   declare py="" toolchain_py
-  toolchain_py="$(compgen -G '/opt/mongodbtoolchain/v*/bin/python3' | sort -V | tail -n1)" || true
-  if [ -n "$toolchain_py" ] && [ -x "$toolchain_py" ]; then
-    py="$toolchain_py"
-  elif command -v python3 >/dev/null 2>&1; then
-    py="$(command -v python3)"
-  elif command -v python >/dev/null 2>&1; then
-    py="$(command -v python)"
+  if [ -n "${VIRTUAL_ENV:-}" ] && [ -x "$VIRTUAL_ENV/bin/python" ]; then
+    py="$VIRTUAL_ENV/bin/python"
+  else
+    toolchain_py="$(compgen -G '/opt/mongodbtoolchain/v*/bin/python3' | sort -V | tail -n1)" || true
+    if [ -n "$toolchain_py" ] && [ -x "$toolchain_py" ]; then
+      py="$toolchain_py"
+    elif command -v python3 >/dev/null 2>&1; then
+      py="$(command -v python3)"
+    elif command -v python >/dev/null 2>&1; then
+      py="$(command -v python)"
+    fi
   fi
 
   # None of these is reliably on PATH in a fresh shell. ~/.local/bin is where uv's
--- a/.evergreen/ensure-uv.sh
+++ b/.evergreen/ensure-uv.sh
@@ -109,16 +109,26 @@ _ensure_uv_install() {
   declare py="${1:?}" venv_dir="${2:?}" log="${3:?}"
 
   if "$py" -m pip --version >/dev/null 2>&1; then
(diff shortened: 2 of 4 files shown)


.evergreen/tests/test-ensure-uv.sh
```diff
@@ -0,0 +1,147 @@
...
+  # Drop any PATH entry that already holds a uv, so ensure_uv has to install its
+  # own rather than reusing whatever the host ships.
+  local cleaned="" p
+  local IFS=":"
+  for p in $PATH; do
+    [ -n "$p" ] || continue
+    if [ -x "$p/uv" ] || [ -x "$p/uv.exe" ]; then continue; fi
+    cleaned="${cleaned:+${cleaned}:}$p"
+  done
+  export PATH="$cleaned"
```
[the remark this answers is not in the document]
blink1073: Addressed: reset_env now shadows any preinstalled uv with a failing stub on PATH (instead of dropping the whole PATH entry), so interpreters co-located with uv (e.g. uv and python3 under /usr/local/bin) stay available. Bash caches the first uv location it finds, so ensure_uv kept resolving to the stub after installing a fresh uv and wrongly failed; I added `hash -r` after each install so ensure_uv re-scans PATH. The regression test passes.

.evergreen/ensure-uv.sh
```diff
@@ -104,21 +104,34 @@ _ensure_uv_add_user_bin() {
 # - The docker test images install a deadsnakes python with venv but no pip, so the
 #   venv covers them as well.
 #
+# Keep the venv fallback. The legacy KMS VMs still need it, and it is the
+# backstop for any host where the pip path cannot install uv.
```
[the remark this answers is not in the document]
blink1073: Addressed: the install-path doc now describes the real order -- install into the active env when possible, otherwise pip install --user, then the fallback venv -- instead of claiming the first attempt is always `pip install --user` and that active venvs only succeed via the fallback.

.evergreen/docker/ubuntu22.04/Dockerfile
```diff
@@ -25,6 +25,7 @@ RUN export DEBIAN_FRONTEND=noninteractive \
 && apt-get -qq update \
 && apt-get -qq -y install $PYTHON $PYTHON-venv < /dev/null > /dev/null \
 && ln -sf /usr/bin/$PYTHON /usr/bin/python3 \
+&& $PYTHON -m ensurepip < /dev/null > /dev/null 2>&1 \
```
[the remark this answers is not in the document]
blink1073: This Dockerfile installs `python3.11` from the `deadsnakes/ppa`, not the Ubuntu system interpreter (3.10) that Debian/Ubuntu patch to disable `ensurepip`. Deadsnakes builds ship `ensurepip` enabled, so `python3.11 -m ensurepip` works and installs pip into that interpreter; there is also no `python3.11-pip` apt package for it. The citations apply to the Debian 11 KMS VMs' `system` python (the reason those need the venv fallback), not to this image. Supporting evidence: these images have long relied on `ensure_uv`'s venv fallback, which bootstraps pip by calling `ensurepip` — if `ensurepip` were disabled there the venv fallback would fail and the current docker tests would already be red.

.evergreen/tests/test-ensure-uv.sh
```diff
@@ -0,0 +1,148 @@
...
+  mkdir -p "$stub/venv"
+  printf 'raise ImportError("venv disabled for test")\n' >"$stub/venv/__init__.py"
+  echo "Testing ensure_uv without a venv module ..."
+  (
+    reset_env
+    export PYTHONPATH="$stub"
+    # Use the pre-resolved interpreter rather than a bare `python3` from PATH:
+    # reset_env drops entries holding uv, and on hosts where uv and python3 are
+    # co-located (e.g. /usr/local/bin) that would hide the interpreter this check
+    # needs.
```
eramongodb: Is this comment still accurate? `reset_env` itself doesn't invoke `ensure_uv`. Per lines 67-68:

> Shadow preinstalled uv with a failing stub so ensure_uv has to install its own, rather than dropping the PATH entry and hiding tools co-located with uv.

Suggest simplifying/stripping/cleaning-up doc comments to reduce them to only what is essential-to-know for future maintenance. It is difficult to validate the accuracy of these doc comments given how much detail they currently contain.

.evergreen/tests/test-remote-kms-provisioning.sh
eramongodb: No longer accurate? Active venv should now be the default rather than the fallback.

.evergreen/ensure-uv.sh
eramongodb: Two -> Three: active venv, then `pip --user`, then fallback venv.

.evergreen/docker/ubuntu22.04/Dockerfile
```diff
@@ -25,6 +25,10 @@ RUN export DEBIAN_FRONTEND=noninteractive \
 && apt-get -qq update \
 && apt-get -qq -y install $PYTHON $PYTHON-venv < /dev/null > /dev/null \
 && ln -sf /usr/bin/$PYTHON /usr/bin/python3 \
+&& ( $PYTHON -m ensurepip < /dev/null > /dev/null 2>&1 \
+     || ( curl -sSfL https://bootstrap.pypa.io/get-pip.py -o /tmp/get-pip.py \
```
eramongodb:
```suggestion
     || ( curl --retry 5 -sSfL https://bootstrap.pypa.io/get-pip.py -o /tmp/get-pip.py \
```
To (try to) avoid spurious network failures.

.evergreen/ensure-uv.sh
eramongodb: Update docs to account for active venv > `pip --user` > fallback venv precedence (+ ensure this order is consistent across the PR).

.evergreen/tests/test-ensure-uv.sh
```diff
@@ -0,0 +1,148 @@
...
+    echo "test-ensure-uv.sh: unsupported platform; skipping."
+    make -C "$DRIVERS_TOOLS" test
+    exit 0
+    ;;
+esac
+
+# ensure_uv only uses a Python 3.8+ interpreter, so build its test venv with one.
+# On RHEL 8 the system python3 is 3.6; fall back to the toolchain when needed.
+PY_BIN=""
+for c in python3 $(compgen -G '/opt/mongodbtoolchain/v*/bin/python3' | sort -Vr) python; do
```
eramongodb:
```suggestion
for c in $(compgen -G '/opt/mongodbtoolchain/v*/bin/python3' | sort -Vr) python3 python; do
```
Reorder to follow `ensure_uv` precedence: [ensure-uv.sh:209-218](https://github.com/blink1073/drivers-evergreen-tools/blob/56e2a0a28616e22a00432c6a13abcdbe6237b6c1/.evergreen/ensure-uv.sh#L209-L218)

eramongodb (changes requested):

.evergreen/docker/ubuntu22.04/Dockerfile
blink1073: Fixed

.evergreen/tests/test-ensure-uv.sh
blink1073: I tightened up the comments

.evergreen/tests/test-remote-kms-provisioning.sh
blink1073: Fixed

.evergreen/ensure-uv.sh
blink1073: Done

.evergreen/tests/test-ensure-uv.sh
blink1073: Done

.evergreen/ensure-uv.sh
eramongodb:
```suggestion
# - PATH (~/.local/bin, the active venv's bin dir, pip's `--user` script directory, and a fallback venv's bin dir)
```
One more triplet.

eramongodb (dismissed):
Minor suggestion remaining; otherwise, LGTM.

blink1073:
I added the "current" toolchain the most likely candidate to have an up-to-date python, and added tests specifically for `windows-vsCurrent` and `windows-2022-latest` to prevent regressions on those hosts.

.evergreen/ensure-uv.sh
```diff
@@ -176,19 +208,35 @@ ensure_uv() {
...
+    fi
+  fi
+  if [ -z "$py" ]; then
+    case "${OSTYPE:-}" in
+    cygwin) current_py="C:/python/Current/python.exe" ;;
+    darwin*) current_py="/Library/Frameworks/Python.Framework/Versions/Current/bin/python3" ;;
+    *) current_py="/opt/python/Current/bin/python3" ;;
+    esac
+    if [ -x "$current_py" ] &&
+      "$current_py" -c 'import pip; import sys; sys.exit(0 if sys.version_info >= (3, 9) else 1)' >/dev/null 2>&1; then
```
eramongodb:
```suggestion
      "$current_py" -c 'import pip; import sys; sys.exit(0 if sys.version_info >= (3, 8) else 1)' >/dev/null 2>&1; then
```
uv requires 3.8+, not 3.9+. Further version refinement (e.g. 3.10+) may be subsequently handled by uv.

.evergreen/tests/test-ensure-uv.sh
```diff
@@ -0,0 +1,159 @@
...
+
+SCRIPT_DIR=$(dirname "${BASH_SOURCE[0]}")
+. "$SCRIPT_DIR/../handle-paths.sh"
+
+case "$(uname -s)" in
+  Linux | Darwin)
+    VENV_SUBDIR=bin
+    UV_NAME=uv
+    ;;
+  MINGW* | MSYS* | CYGWIN*)
```
eramongodb:
```suggestion
  CYGWIN*)
```
Remove mingw/msys envs? IIRC none exist in EVG and the current ensure-uv.sh does not special-case them either (cygwin only). Conversely, update ensure-uv.sh so mingw/msys is consistently handled alongside cygwin.

.evergreen/ensure-uv.sh
```diff
@@ -72,61 +72,92 @@ _ensure_uv_add_path() {
...
+  # A native Windows interpreter ends its stdout lines with \r\n; a trailing
+  # \r would corrupt the PATH entry and break bash's lookup of the directory.
+  base="$("${1:?}" -c 'import os, sysconfig; print(sysconfig.get_path("scripts", "nt_user" if os.name == "nt" else "posix_user"))' 2>/dev/null | tr -d '\r')" || return 0
   [ -n "$base" ] || return 0
-  _ensure_uv_add_path "$base/bin"
-  _ensure_uv_add_path "$base/Scripts"
+  if [ "${OSTYPE:-}" = cygwin ]; then
+    # A native Windows interpreter reports a C:\ style path; cygpath it for
+    # PATH. A Cygwin python's POSIX path passes through unchanged.
+    base="$(cygpath -m "$base")"
```
eramongodb:
```suggestion
    base="$(cygpath -m "$base")"
```
Suggest stripping inaccurate comment, [quote](https://cygwin.com/cygwin-ug-net/cygpath.html):

> - `-m, --mixed`: like `--windows`, but with regular slashes (`C:/WINNT`)
> - `-w, --windows`: print Windows form of NAMEs (`C:\WINNT`)

`-m` preserves `C:`-style paths but uses forward slashes instead of backslashes: `C:\example\path` -> `C:/example/path`. Review other calls to `cygpath` and associated comments for consistency/redundancy.

.evergreen/docker/ubuntu22.04/Dockerfile
```diff
@@ -25,6 +25,10 @@ RUN export DEBIAN_FRONTEND=noninteractive \
 && apt-get -qq update \
 && apt-get -qq -y install $PYTHON $PYTHON-venv < /dev/null > /dev/null \
 && ln -sf /usr/bin/$PYTHON /usr/bin/python3 \
+&& ( $PYTHON -m ensurepip < /dev/null > /dev/null 2>&1 \
+     || ( curl --retry 5 -sSfL https://bootstrap.pypa.io/get-pip.py -o /tmp/get-pip.py \
+          && $PYTHON /tmp/get-pip.py < /dev/null > /dev/null 2>&1 \
+          && rm -f /tmp/get-pip.py ) ) \
```
eramongodb: Is this fallback really necessary, given this is an OCI image with explicit PPA above? Would prefer to avoid curl commands without checksum-validation when able (+ get-pip.py doesn't appear to be a stable script that is reliably checksummable either).

eramongodb (changes requested):

.evergreen/ensure-uv.sh
blink1073: Done

.evergreen/tests/test-ensure-uv.sh
blink1073: Done

.evergreen/ensure-uv.sh
blink1073: Done

.evergreen/docker/ubuntu22.04/Dockerfile
blink1073: I removed the fallback

.evergreen/config.yml
```diff
@@ -633,6 +641,16 @@ functions:
...
 pre:
+  # Inline rather than a script because this runs before git.get_project, so the
+  # checkout that would hold the script does not exist yet.
+  - command: shell.exec
+    params:
+      shell: bash
+      script: |
+        if [ "${clear_persisted_uv|}" = "true" ]; then
+          echo "Removing any uv an earlier task left behind, so ensure_uv has to pick an interpreter."
+          rm -f "$HOME/.local/bin/uv"
```
eramongodb, line 652: It's somewhat 🤔❓ to unconditionally modify the user environment in a manner which may break other tasks running on this EVG host, but EVG tasks technically shouldn't make assumptions about the state of the host anyways (including user directories), so I'm left with mixed feelings but no particular suggestion here.

eramongodb (approved):
One last remark, but otherwise LGTM.

.evergreen/config.yml
blink1073: My thoughts:

- The fact that we're writing into $HOME at all is a bit suspect
- We're only doing this on the one host
- The tasks run one at a time, so worst case the next user of the host has to bootstrap it again

